PR #9, opened 2020-06-24
Move docs/acra and docs/themis to top-level

ilammy:
We are going to host documentation on https://docs.cossacklabs.com so having additional `/docs/` in the URL is redundant. Move product docs up one directory layer.

Replace `/docs/themis/...` with just `/themis/...` in all URIs after the documentation directory has been moved.

Enjoy your `git blame` now 😉


ilammy:
@ColliderBang, please confirm that this structure will work with the actual doc server. After that this PR may be merged.

PR #10, opened 2020-07-01
Adjust Themis main page title

ColliderBang:
The item in the configuration is created for display in the header of the site.
Changed the main title to display in the left menu.

### DIFF
--- a/themis/_index.md
+++ b/themis/_index.md
@@ -1,7 +1,8 @@
 ---
 weight: 1
 bookFlatSection: true
-title: "Themis"
+title: "Themis in a nutshell"
+secondTitle: "Themis"
 Menu: "main"
 ---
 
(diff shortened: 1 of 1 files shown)


themis/_index.md
```diff
@@ -1,7 +1,8 @@
 ---
 weight: 1
 bookFlatSection: true
-title: "Themis"
+title: "Themis in a nutshell"
+secondTitle: "Themis"
```
ilammy, line 5: Uh... so, which title goes where? Is `secondTitle` our own extension?

I'd guess, the page title (i.e., the browser window, bookmarks, etc.) will say “Themis in a nutshell” while the menu will say “Themis”? Or will it be reverse?

vixentael: @ColliderBang  if it's ours extension, maybe we should rename `secondTitle` into `menuTitle` or similar?

ilammy: Here's how it looks:

Screen Shot 2020-07-01 at 17 52 14

The `title` item still refers to the page title and the page navigation menu title. The new item `productMenuTitle` sets the title in the product menu (at the top).

PR #11, opened 2020-07-02
Merge master -> staging

vixentael:


ilammy:
It’s better to use real merge commits rather than squashing to merge stuff between long-running branches. It makes it easier to track what changes are included into the branch. It also works better with conflict resolution (though, normally, in the master -> staging workflow there should be no conflicts unless hotfixes are applied, and even then it’s usually clear).

PR #17, opened 2020-08-03
Acra key management

ilammy:
Begin porting Acra documentation to the new platform, starting with key management.

This guide is a compilation of existing docs on the documentation server, some early drafts of documentation for key store v2, and pages from staging version about key rotation. However, since now we are not constrained by the previous limitations, we can introduce more sections and more pages.

There are no major changes to content. I've mostly reformatted it a bit to render more nicely with Hugo, fixed up links, extended some parts to include more information about key store v2. Note that old links to the old documentation are converted to absolute URLs so that they actually work when rendered, and that we know where they lead.

Also, currently Hugo does not render the side menu for Acra (nor the menu leads to its sections). If you're going to try to preview this, you will need to type in URLs manually.

This version is in no way complete. Some pages are missing and I intend to add them before this PR is merged. Some parts of the key store v2 API are not described very well. Since we're probably going to include key store v2 in upcoming Acra release, the documentation is written as if key store v2 is available. However, it is implied that it's new and not a recommended default right now.

Work on this documentation has revealed some glaring holes in key store v2 implementation, which will be fixed later, probably. It just makes me embarrassed and disappointed. All in all, things like this prove that months of work can save days of planning!


acra/keys/inventory.md
```diff
@@ -0,0 +1,99 @@
...
+title: Inventory of Acra keys
+bookCollapseSection: true
+---
+
+# Inventory of Acra keys
+
+There are several types of keys used in Acra:
+
+  - **Storage keypairs.**
+    AcraWriter uses the public key to encrypt data to be stored in the database.
```
vixentael, line 12:
```suggestion
<!-- TODO: describe keypairs in more details: what is keypair, private and public part, introduce "client-side" encryption and "server-side/proxy-side" encryption --> 
    AcraWriter uses the public key to encrypt data to be stored in the database.
```

PR #19, opened 2020-08-11
Use single ACRA_MASTER_KEY for key store v2

ilammy:
Key store v2 has been updated to use a single key store variable. Update the text to use the same variable. Most of the differences are now just simple `--keystore=v1` vs. `v2`. (And a special case for **acra-migrate-keys**.)

This is the librarian sibling of cossacklabs/acra#399 and cossacklabs/acra-enterprise#80.

### DIFF
--- a/acra/keys/operations/generation.md
+++ b/acra/keys/operations/generation.md
@@ -57,7 +57,7 @@ With key store version 2 you will need to use `acra-keymaker --keystore=v2`.
 
 {{< hint info >}}
 **Note:**
-If you are using Acra 0.85 or earier,
+If you are using Acra 0.85 or earlier,
 please omit the `--keystore` parameter here and onward.
 {{< /hint >}}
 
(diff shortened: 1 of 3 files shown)


acra/keys/operations/generation.md
```diff
@@ -48,20 +48,18 @@ such as hardware security modules (HSM) and key management services (KMS).
...
-```shell
-acra-keymaker --generate_master_key=master_encryption.key
-acra-keymaker --generate_master_key=master_signature.key
-
-export ACRA_MASTER_ENCRYPTION_KEY=$(cat master_encryption.key | base64)
-export ACRA_MASTER_SIGNATURE_KEY=$(cat master_signature.key | base64)
-```
+{{< hint info >}}
+**Note:**
+If you are using Acra 0.85 or earier,
```
shadinua, line 60:
```suggestion
If you are using Acra 0.85 or earlier,
```

acra/keys/operations/generation.md
```diff
@@ -142,20 +135,18 @@ Generate a second set of master keys for AcraConnector in the same way,
...
```
shadinua, line 147:
```suggestion
If you are using Acra 0.85 or earlier,
```

PR #20, opened 2020-08-14
Update installation and import instructions for ObjC and Swift

vixentael:
Since Themis 0.13.2 release, we've switched back to use `themis.framework` instead of `objcthemis.framework`. Update Carthage installation instructions and ObjC usage instructions accordingly.

### DIFF
--- a/themis/languages/objc/features.md
+++ b/themis/languages/objc/features.md
@@ -13,7 +13,7 @@ it is ready to use in your application!
 In order to use ObjCThemis, you need to import its headers:
 
 ```objc
-#import <objcthemis/objcthemis.h>
+@import themis;
 ```
 
 ---
--- a/themis/languages/objc/features.md
+++ b/themis/languages/objc/features.md
@@ -10,7 +10,7 @@ it is ready to use in your application!
 
 ## Using Themis
 
-In order to use ObjCThemis, you need to import its headers:
+In order to use ObjCThemis, you need to import it:
 
 ```objc
 @import themis;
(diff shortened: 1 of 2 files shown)


themis/languages/objc/features.md
```diff
@@ -13,7 +13,7 @@ it is ready to use in your application!
 In order to use ObjCThemis, you need to import its headers:
```
ilammy, line 13:
```suggestion
In order to use ObjCThemis, you need to import it:
```
Since there are no more headers to import or include.

themis/languages/objc/installation.md
```diff
@@ -64,22 +64,12 @@ Themis is also available via [**Carthage**](https://github.com/Carthage/Carthage
...
-You should integrate *this* framework into your applications.
-
-Currently, for Themis 0.13,
-Carthage will build both new "objcthemis.framework" and old "themis.framework"
-to allow existing users to migrate.
-Do not use "themis.framework" in new applications
-and update existing applications to use "objcthemis.framework" instead.
-
-The old "themis.framework" is scheduled to be **removed** in Themis 0.14.
+You should integrate only *themis.framework* into your applications. Do not integrate *openssl.framework* manually, as it is statically linked with Themis (Carthage handles linking.).
```
ilammy, line 72: Well, it probably won't hurt if openssl.framework is there, but maybe App Store will reject loose static frameworks in bundle or whatever.

vixentael: less is more

PR #25, opened 2020-09-07
Themis specification: keys and keypairs

ilammy:
Start describing Themis cryptography and data structure design in a series of technical specifications. Hopefully, this will make reimplementing Themis in different platforms easier as well as simplifies security audit and analysis. It may also be beneficial to new developers to serve as a map for the codebase (or rather, as a textbook which you apply to the codebase).

It will take a while to write so unfinished sections get a notice instead of content. The pages are already available so that the hyperlinks work right away.

<sub>Yeah, instead of pointlessly rewriting Themis in Rust, I started pointlessly rewriting it in English. Feel free to join.</sub>

### DIFF
--- a/themis/spec/asymmetric-keypairs/_index.md
+++ b/themis/spec/asymmetric-keypairs/_index.md
@@ -5,13 +5,6 @@ title:  Asymmetric keypairs
 
 # Asymmetric keypairs
 
-{{< hint info >}}
-This part of Themis documentation is currently under development.
-Please come back later.
-
-Meanwhile you can read our [key management guide](/themis/crypto-theory/key-management/).
-{{< /hint >}}
-
 Asymmetric keypairs are used by [**Secure Message**](../secure-message/)
 and [**Secure Session**](../secure-session/) cryptosystems.
 
--- a/themis/spec/asymmetric-keypairs/_index.md
+++ b/themis/spec/asymmetric-keypairs/_index.md
@@ -19,3 +19,9 @@ Currently, Themis supports the following asymmetric keys:
 
 - [EC keys](ecc/) used by Secure Message and Secure Session
 - [RSA keys](rsa/) supported only by Secure Message
+
+EC keys are preferred over RSA because EC keys can be shorter while providing comparable security guarantees.
+Moreover, elliptic cryptography is more computationally efficient for most use cases.
+<!-- TODO: provide a link to Secure Message section which benchmarks the keys -->
+However, you may still need RSA in order to comply with certain regulations.
+Read more about compliance on the [Regulations](/themis/regulations/) page.
--- a/themis/spec/asymmetric-keypairs/_index.md
+++ b/themis/spec/asymmetric-keypairs/_index.md
@@ -1,0 +1,13 @@
+---
+weight: 81
+title:  Asymmetric keypairs
+---
+
+# Asymmetric keypairs
+
+{{< hint info >}}
+This part of Themis documentation is currently under development.
+Please come back later.
+
+Meanwhile you can read our [key management guide](/themis/crypto-theory/key-management/).
+{{< /hint >}}
--- a/themis/spec/asymmetric-keypairs/_index.md
+++ b/themis/spec/asymmetric-keypairs/_index.md
@@ -11,3 +11,18 @@ Please come back later.
 
 Meanwhile you can read our [key management guide](/themis/crypto-theory/key-management/).
 {{< /hint >}}
+
+Asymmetric keypairs are used by [**Secure Message**](../secure-message/)
+and [**Secure Session**](../secure-session/) cryptosystems.
+
+Asymmetric cryptography imposes certain mathematical requirements on the structure of the keys.
+To preserve the format and content of the keys intact,
+all asymmetric keys are stored in [Soter containers](../common/#soter-container)
+which provide key type identification and prevent accidental corruption.
+
+## Asymmetric key kinds
+
+Currently, Themis supports the following asymmetric keys:
+
+- [EC keys](ecc/) used by Secure Message and Secure Session
+- [RSA keys](rsa/) supported only by Secure Message
--- a/themis/spec/asymmetric-keypairs/ecc.md
+++ b/themis/spec/asymmetric-keypairs/ecc.md
@@ -5,13 +5,6 @@ title:  EC keys
 
 # EC keys
 
-{{< hint info >}}
-This part of Themis documentation is currently under development.
-Please come back later.
-
-Meanwhile, you may read our [key management guide](/themis/crypto-theory/key-management/).
-{{< /hint >}}
-
 In *elliptic curve cryptography* (ECC), keys are defined by a number of parameters.
 **Domain parameters** describe a particular curve used for operations.
 It is critical to use a good curve with secure properties,
--- a/themis/spec/asymmetric-keypairs/ecc.md
+++ b/themis/spec/asymmetric-keypairs/ecc.md
@@ -167,7 +167,7 @@ The recommended way to generate EC keypairs is to use a cryptography library.
 See also our [key management guide](/themis/crypto-theory/key-management/)
 for advice on securing the keys after you have generated them.
 
-Here is a snippet in Go, illustrating how to generate a good EC keypair:
+Here is a snippet in Go, illustrating how to generate EC keypair using P256 curve:
 
 ```go
 package main
--- a/themis/spec/asymmetric-keypairs/ecc.md
+++ b/themis/spec/asymmetric-keypairs/ecc.md
@@ -120,7 +120,8 @@ where you can see the data fields:
 | k       | 0x0D   | `D2 80 F9...` (32 bytes) | the **k** value |
 
 You can verify that the keys match each other on the P-256 curve
-and that the **y** coordinate value is indeed odd:
+and that the **y** coordinate value is indeed odd
+using the following snippet in Go:
 
 ```go
 package main
@@ -166,7 +167,7 @@ The recommended way to generate EC keypairs is to use a cryptography library.
 See also our [key management guide](/themis/crypto-theory/key-management/)
 for advice on securing the keys after you have generated them.
 
-Here is a snippet illustrating how to generate a good EC keypair:
+Here is a snippet in Go, illustrating how to generate a good EC keypair:
 
 ```go
 package main
--- a/themis/spec/asymmetric-keypairs/ecc.md
+++ b/themis/spec/asymmetric-keypairs/ecc.md
@@ -162,7 +162,7 @@ func main() {
 
 ## Generating EC keys
 
-The recommended way to generate EC keypairs is to use a cryptography library.
+The recommended way to generate EC keypairs is to use a cryptographic library.
 
 See also our [key management guide](/themis/crypto-theory/key-management/)
 for advice on securing the keys after you have generated them.
@@ -173,8 +173,8 @@ Here is a snippet in Go, illustrating how to generate a good EC keypair:
 package main
 
 import (
-	"crypto/elliptic" // Use a cryptography library.
-	"crypto/rand"     // Use a cryptograpic RNG.
+	"crypto/elliptic" // Use a cryptographic library.
+	"crypto/rand"     // Use a cryptographic RNG.
 	"fmt"
 )
 
--- a/themis/spec/asymmetric-keypairs/ecc.md
+++ b/themis/spec/asymmetric-keypairs/ecc.md
@@ -11,3 +11,221 @@ Please come back later.
 
 Meanwhile, you may read our [key management guide](/themis/crypto-theory/key-management/).
 {{< /hint >}}
+
+In *elliptic curve cryptography* (ECC), keys are defined by a number of parameters.
+**Domain parameters** describe a particular curve used for operations.
+It is critical to use a good curve with secure properties,
+therefore usually one of the well-known respectable curves is chosen.
+
+As of Themis 0.13 released in 2020,
+the following elliptic curves are supported:
+
+  - NIST P-256, aka secp256r1, aka prime256v1 **(current default)**
+  - NIST P-384, aka secp384r1
+  - NIST P-521, aka secp521r1
+
+All of them have been recommended by the NSA Suite B.
+
+{{< hint info >}}
+Migration to more recent recommendations of the CNSA Suite is pending.
+{{< /hint >}}
+
+As for variable parameters of the key,
+a scalar integer **k** is used as a private key.
+A point on the curve **P**, obtained by multiplying the base point of the curve by **k**, is the corresponding public key.
+
+## Layout
+
+EC keys are stored in [Soter containers](../../common/#soter-container).
+The tags for the supported curves are as follows:
+
+| Curve | Public key tag | Private key tag |
+| ----- | -------------- | --------------- |
+| P-256 | `UEC2`         | `REC2`          |
+| P-384 | `UEC3`         | `REC3`          |
+| P-521 | `UEC5`         | `REC5`          |
+
+The payload of Soter container stores the key data encoded according to ANSI X9.62.
+All fields are stored in big-endian (network) byte order.
+
+Public key layout (for the P-256 curve):
+
+     0 1                                                               33
+    +-+-------+-------+-------+-------+-------+-------+-------+-------+
+    |z|                               x                               |
+    +-+-------+-------+-------+-------+-------+-------+-------+-------+
+
+where
+
+  - **z** (1 byte) is an indicator of the y coordinate value
+  - **x** (32 bytes) is the x coordinate of the public point **P**
+
+**z** is equal to `02` if the y coordinate of the public point is even, or `03` if it is odd.
+No other values are permitted.
+This is known as *compressed representation* of an elliptic curve point,
+possible due to the fact that there are at most two points on a curve with a given x coordinate.
+
+Private key layout (for the P-256 curve):
+
+     0 1                                                               33
+    +-+-------+-------+-------+-------+-------+-------+-------+-------+
+    |0|                               k                               |
+    +-+-------+-------+-------+-------+-------+-------+-------+-------+
+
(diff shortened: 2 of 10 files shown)


iamnotacake:
Things I've seen in this PR so far look good and well explained. Of course, one still have to understand the math behind EC & RSA to implement them properly. Or just use the existing libraries. I guess you gonna add SMessage & SSession specs later in this same PR.

themis/spec/common.md
```diff
@@ -11,3 +11,109 @@ Please come back later.
...
+where you can see the data fields:
+
+| Field   | Offset | Data          | Meaning |
+| ------- | ------ | ------------- | ------- |
+| tag     | 0x00   | `55 45 43 32` | `UEC2` – EC public key for the P-256 curve |
+| size    | 0x04   | `00 00 00 2D` | 45 bytes: 12 bytes of header, 33 bytes of payload |
+| crc     | 0x08   | `6C D5 6E F8` | container checksum |
+| payload | 0x0C   | `03 A4 B1...` | data payload: public key parameters |
+
+Here is a code snippet illustrating computation of Soter container checksum:
```
storojs72, line 76: I would add that this is a Go's programming language snippet. There are other similar places:
- https://github.com/cossacklabs/product-docs/pull/25/commits/63faf3d3b0b7355f4e5b4eb0cf4ccb953cc40be3#diff-90c4f24e395ae483f38d1408d0cf47adR54
- https://github.com/cossacklabs/product-docs/pull/25/commits/1753920b0274fc8c0401660e2fae093a3e2ed245#diff-fe4809a11f4d9d1bfaa450caf9723471R130
- https://github.com/cossacklabs/product-docs/pull/25/commits/7d8fcae2e93c889cb3b7819e2d5fb4a643c021d4#diff-14ad7a0c001539a2ca8334e3e21fb1e0R231
- https://github.com/cossacklabs/product-docs/pull/25/commits/7d8fcae2e93c889cb3b7819e2d5fb4a643c021d4#diff-14ad7a0c001539a2ca8334e3e21fb1e0R355

themis/spec/symmetric-keys.md
```diff
@@ -12,4 +12,95 @@ Please come back later.
...
+		panic("failed to generate a key")
+	}
+	fmt.Printf("✅ symmetric key: %x\n", key)
+}
+```
+
 ## Passphrases
+
+Alternatively, Secure Cell can work with passphrases.
+From the cryptographic point of view, a passphrase is essentially a weird symmetric key.
```
storojs72, line 75: I would write not "weird", but "weak". But its just IMO

themis/spec/asymmetric-keypairs/_index.md
```diff
@@ -11,3 +11,18 @@ Please come back later.
...
+To preserve the format and content of the keys intact,
+all asymmetric keys are stored in [Soter containers](../common/#soter-container)
+which provide key type identification and prevent accidental corruption.
+
+## Asymmetric key kinds
+
+Currently, Themis supports the following asymmetric keys:
+
+- [EC keys](ecc/) used by Secure Message and Secure Session
+- [RSA keys](rsa/) supported only by Secure Message
```
storojs72, line 21: I would explain here, why we use ECC keys by default. The reason is lower key sizes of ECC keys compared to RSA keys having same levels of security: https://www.globalsign.com/en/blog/elliptic-curve-cryptography (we can even add a table from that blogpost). Plus ECC crypto has lower CPU consumption (https://crypto.stackexchange.com/questions/1190/why-is-elliptic-curve-cryptography-not-widely-used-compared-to-rsa). The only advantage of RSA is that it is better established historically

themis/spec/asymmetric-keypairs/rsa.md
```diff
@@ -11,3 +11,442 @@ Please come back later.
...
+		fmt.Println("✅ qp⋅q ≡ 1 (mod p)")
+	} else {
+		fmt.Println("❌ qp⋅q ≡ 1 (mod p)")
+	}
+}
+```
+
+## Generating RSA keys
+
+The recommended way to generate RSA keypairs is to use a cryptography library.
```
storojs72, line 343: seems it should be "cryptographic library"

storojs72 (commented):
Very valuable specification. I have just some minor suggestions. Thank you @ilammy. Looking at your PRs arrangement just requires me to do the same and dedicate more time on proper git commit and merge practice

themis/spec/common.md
ilammy: > I would add that this is a Go's programming language snippet

Sure. It won't hurt to be explicit.

themis/spec/symmetric-keys.md
ilammy: > I would write not "weird", but "weak" [symmetric key].

Mmm... It's not like passphrases are weak _per se_: given a sufficient length and randomness, they can be effectively as good as the keys. I guess I wanted to communicate here that passphrases have lower entropy per byte than keys, that's why they are "worse".

I'll think of a better wording. At the moment, I'm out of ideas. @vixentael?

themis/spec/asymmetric-keypairs/_index.md
ilammy: > I would explain here, why we use ECC keys by default.

Yeah, makes sense. I'm not sure if we should go into much detail here, but it will be good to note that ECC is preferred to RSA and we support RSA mostly for compliance reasons. Secure Message page could go in more details on what effects the key choice has on the practical performance of the underlying algorithms.

themis/spec/asymmetric-keypairs/rsa.md
ilammy: > seems it should be "cryptographic library"

IDK :joy: During editing I've flipped it between "cryptography" and "cryptographic" several times. Unless there are strong opinions, I guess I'll go ahead to grep the variants in our own docs and just use the most popular one.

ilammy:
@storojs72, thanks for review. Your opinion is very important and valuable here.

> requires me to do the same and dedicate more time on proper git commit and merge practice

Oh crap, you have uncovered my secret agenda! :scream:

ilammy:
@iamnotacake,

> Of course, one still have to understand the math behind EC & RSA to implement them properly. Or just use the existing libraries.

I don't intend this spec as a primer on cryptographic algorithms so they won't be explained in too much detail. The preferred way is still to avoid rolling your own crypto and just use libraries maintained by experts. It's just happens so that there aren't many libraries around which can generate keys in the format Themis wants them, so we go in more detail in this case, explaining the components of the key containers and how they can be computed (but _without_ explaining how to multiply EC points or test numbers for primality).

> I guess you gonna add SMessage & SSession specs later in this same PR.

Not in this PR, but in later ones, yeah. As soon as the algorithms are understood. I'd like to keep PRs reviewable so they shouldn't contain too much content.

themis/spec/asymmetric-keypairs/rsa.md
ilammy: "Cryptographic" it is.

```
$ git grep -hoi 'crypto[a-z]*' | sort | uniq -c
    351 crypto
      3 Crypto
      5 CRYPTO
      2 cryptoalgorithm
      5 cryptoalgorithms
      1 cryptoanalysis
      2 cryptographer
      3 cryptographers
    135 cryptographic
      4 Cryptographic
     12 cryptographically
     90 cryptography
     15 Cryptography
      1 cryptograpic
      1 cryptomuseum
      5 cryptoperiod
      1 cryptoperiods
     81 cryptosystem
      3 Cryptosystem
    311 cryptosystems
      5 Cryptosystems
```

themis/spec/common.md
ilammy: I've added a mention of Go before each snippet.

themis/spec/symmetric-keys.md
ilammy: Well, if that's about entropy then I've just written *that* as is, without calling passphrases weird. Please give it a look.

themis/spec/asymmetric-keypairs/_index.md
ilammy: I've added a short blurb about EC vs RSA keys, which hopefully provides a bit more context on why Themis supports multiple types of keys and which is the preferred one.

themis/spec/asymmetric-keypairs/rsa.md
vixentael: `cryptography library` is more common used term, better for search

https://trends.google.com/trends/explore?q=cryptography%20library,cryptographic%20library
Screenshot 2020-09-11 at 14 00 15

ilammy: > cryptography library is more common used term, better for search

Okie dokie, reverted.

ilammy:
Are there any more blockers for this PR?

vixentael:
> Are there any more blockers for this PR?

Please don't merge until my review approval

themis/spec/asymmetric-keypairs/ecc.md
```diff
@@ -0,0 +1,225 @@
+---
+weight: 10
+title:  EC keys
+---
+
+# EC keys
+
+In *elliptic curve cryptography* (ECC), keys are defined by a number of parameters.
+**Domain parameters** describe a particular curve used for operations.
+It is critical to use a good curve with secure properties,
```
vixentael, line 10:
```suggestion
It is critical to use a "good" curve with validated security properties,
```

themis/spec/asymmetric-keypairs/ecc.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+## Generating EC keys
+
+The recommended way to generate EC keypairs is to use a cryptography library.
+
+See also our [key management guide](/themis/crypto-theory/key-management/)
+for advice on securing the keys after you have generated them.
+
+Here is a snippet in Go, illustrating how to generate a good EC keypair:
```
vixentael, line 170:
```suggestion
Here is a snippet in Go, illustrating how to generate EC keypair using P256 curve:
```

themis/spec/symmetric-keys.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+## Passphrases
+
+Alternatively, Secure Cell can work with passphrases.
+From the cryptographic point of view, a passphrase is essentially a symmetric key with lower entropy per byte.
+Themis encodes textual passphrases in UTF-8 and uses the resulting byte string.
+
+Since passphrases are typically shorter than keys and have less entropy (randomness),
+Secure Cell uses additional KDF rounds to preprocess passphrase data into an encryption key.
```
vixentael, line 79:
```suggestion
Secure Cell uses additional PBKDF2 rounds to preprocess passphrase data into an encryption key.
```

let's say somewhere that we use PBKDF2?

themis/spec/symmetric-keys.md
```diff
@@ -0,0 +1,99 @@
...
+
+{{< hint danger >}}
+**Warning:**
+As for definitely **not recommended** approaches, avoid letting users choose their passphrases.
+Humans are rarely capable of generating cryptographically secure random sentences.
+{{< /hint >}}
+
+There are no strict recommendations for a passphrase length since it mostly depends on human memory capabilities,
+but typically this is about 80 bits for offline passphrases and at least 128 bits for online use.
+Obviously, the longer the better.
```
vixentael, line 99:
```suggestion
[NIST SP 800-63](https://www.nist.gov/itl/tig/projects/special-publication-800-63) suggests using 8 characters as minimum, and up to 64 characters for user-generated passphrases (see [NIST SP 800-63b, 5.1.1 Memorized Secrets](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63b.pdf)).
```
```

File: themis/spec/symmetric-keys.md [hunk 9]
```diff
@@ -0,0 +1,99 @@
...
+### Generating passphrases
+
+The recommended approach to generating passphrases is to generate a symmetric key of an adequate length,
+then render it into a mnemonic which is easier for humans to remember.
+For example, you can use a list of 2048 words to convert a 128-bit key into a sequence of 12 words.
+
+{{< hint danger >}}
+**Warning:**
+As for definitely **not recommended** approaches, avoid letting users choose their passphrases.
+Humans are rarely capable of generating cryptographically secure random sentences.
```
vixentael, line 94:
```suggestion
As for definitely **not recommended** approaches, avoid letting users choose their cryptographic keys, as humans are rarely capable of generating cryptographically secure random sentences. 
```
I don't know where "avoid letting users choose their passphrases" is coming from. Memorizable secrets are foundation for many security controls, NIST guides have certain rules for handling memorizable secrets.

These sentences should be rewritten or removed.

themis/spec/asymmetric-keypairs/rsa.md
```diff
@@ -0,0 +1,445 @@
...
+
+See also our [key management guide](/themis/crypto-theory/key-management/)
+for advice on securing the keys after you have generated them.
+
+Here is a snippet in Go, illustrating how to generate a good RSA keypair:
```
vixentael, line 348:
```suggestion
Here is a snippet in Go, illustrating how to generate a long RSA keypair:
```

themis/spec/asymmetric-keypairs/ecc.md
ilammy: Why quotes then?
```suggestion
It is critical to use a curve with validated security properties,
```

themis/spec/symmetric-keys.md
ilammy: I've explicitly avoided mentioning PBKDF2 here since it's only an overview of keys and passphrases, not Secure Cell operation. I believe, intricacies of the KDF choice belong to Secure Cell description specifically. For one, it's not only PBKDF2 there, ZRTP-style KDF is also thrown into the mix. It will be easier to have KDF description in one place so that when a new KDF is added to Secure Cell then it's not necessary to locate and edit a bunch of other places as well. Passphrase content does not really depend on the specific KDF in use.

themis/spec/asymmetric-keypairs/rsa.md
ilammy: The point here isn't to have a long keypair. If you're concerned, how about we drop the word completely?
```suggestion
Here is a snippet in Go, illustrating how to generate an RSA keypair:
```

themis/spec/symmetric-keys.md
ilammy: Well... It's contestable whether guidelines for *authentication* apply to secrets used to secure data. But okay, a reference is better than no reference.

themis/spec/symmetric-keys.md
ilammy: > I don't know where "avoid letting users choose their passphrases" is coming from.

From my head, obviously.

The stress here is on *"cryptographically secure"* (i.e., with comparable security to keys). Sure, humans are choosing their passphrases and ultimately it comes down to something kept in the head. However, without training and education, typical people are unlikely to choose a strong passphrase either, just like they are usually not able to generate keys of acceptable quality from their head.

Current Secure Cell KDF parameters are tweaked for *strong* passphrases. The slowdown of KDF contributes only about a dozen additional entropy bits and the passphrase is responsible for the rest. We recommend 256-bit keys, so my overall understanding is that the passphrases should have comparable strength. Or, well, at least not expect *majority* of entropy to come from the KDF alone and allow users to choose 4-digit PINs, claiming them to be as secure as the keys generated by Themis. For that, we'd need way more KDF rounds.

Anyways, if this is controversial and we don't have any advice to give here then I'm for removing this paragraph.

themis/spec/symmetric-keys.md
vixentael: If not mention PBKD2 explicitly, maybe let's link to a page where we describe usage of PBKDF2?

themis/spec/asymmetric-keypairs/rsa.md
vixentael: I don't think we should be using "good" as description of cryptographically-secure keys/keypairs in official docs, that's why I suggest other options.

themis/spec/symmetric-keys.md
vixentael: > then I'm for removing this paragraph.

or let's comment it out? maybe future-us will get back here and update this piece

themis/spec/symmetric-keys.md
ilammy: > maybe let's link to a page where we describe usage of PBKDF2?

As soon as it there. My idea was to cross-link Secure Cell spec with key/passphrase spec once it's actually ready.

themis/spec/symmetric-keys.md
ilammy: > or let's comment it out? maybe future-us will get back here and update this piece

I'm not a fan of leaving swathes of commented out code, or documentation. I thought about replacing it with a `<!-- TODO -->`, linking to this discussion. But after further consideration I don't really see anything of value to be put here. We already give an idea of how a passphrase can be generated and give links to NIST recommendations on the topic. I guess that should be enough. If future editors are really interested, they can `git blame` the file.

vixentael: okay please do.

themis/spec/symmetric-keys.md
vixentael: We already have mentions of PBKDF2 in docs. I suggest to put a link on Secure Cell docs. https://docs.cossacklabs.com/themis/crypto-theory/cryptosystems/secure-cell/#key-derivation-functions

vixentael (approved):

themis/spec/symmetric-keys.md
ilammy: > We already have mentions of PBKDF2 in docs. I suggest to put a link on Secure Cell docs.
> https://docs.cossacklabs.com/themis/crypto-theory/cryptosystems/secure-cell/#key-derivation-functions

Okay, I've added links to this section.

vixentael (approved):

vixentael:
Merge 'em all!

PR #31, opened 2020-11-27
Secure Cell layout

ilammy:
Continuing from #30, describe data layout of Secure Cell in detail. Explain how the authentication tokens are structured, for both symmetric key and passphrase cases, and describe the Soter algorithm descriptor too which is an integral part of the authentication token (and used by some other cryptosystems as well).

Also, provide dissected example data to illustrate the layout described above. It is possible to cross-check the data using the interactive simulator or command-line utilities.

This may be a bit *overly* detailed description, but hey, “if you have a point to make, hit it once, then again, then a third time”. Though please feel free to restrain me. By now, I should have a well-established reputation of graphomania, evidently.

Again, there are a lot of people on the reviewer list, but I'd like @vixentael and @iamnotacake to pay the most attention here.

### DIFF
--- a/themis/spec/secure-cell.md
+++ b/themis/spec/secure-cell.md
@@ -306,7 +306,7 @@ This results in the following descriptors currently being in use for symmetric k
 | Encryption algorithm       | Algorithm ID |
 | -------------------------- | ------------ |
 | AES-256-GCM **(default)**  | `0x40010100` |
-| AES-196-GCM                | `0x400100C0` |
+| AES-192-GCM                | `0x400100C0` |
 | AES-128-GCM (_deprecated_) | `0x40010080` |
 
 Refer to the [algorithm descriptor overview](../common/#soter-symmetric-algorithm-descriptor) for details.
@@ -359,7 +359,7 @@ This results in the following **algorithm ID** values currently being in use for
 | Encryption algorithm       | Algorithm ID |
 | -------------------------- | ------------ |
 | AES-256-GCM **(default)**  | `0x41010100` |
-| AES-196-GCM                | `0x410100C0` |
+| AES-192-GCM                | `0x410100C0` |
 | AES-128-GCM (_deprecated_) | `0x41010080` |
 
 Refer to the [algorithm descriptor overview](../common/#soter-symmetric-algorithm-descriptor) for details.
--- a/themis/spec/secure-cell.md
+++ b/themis/spec/secure-cell.md
@@ -220,3 +220,166 @@ AES-GCM algorithm has a limit on how much data can be safely encrypted using the
 before cryptanalysis attacks become practical.
 KDF mixes in user-provided associated data into the derived key,
 randomising it, and reducing effective key reuse even when the same secret is used with Secure Cell.
+
+## Layout
+
+The two main parts of a Secure Cell are:
+
+  - **encrypted data** which is self-explanatory, and
+  - **authentication token** which keeps all auxiliary metadata
+
+Their arrangement is defined by Secure Cell operation mode.
+
+In Seal mode the authentication token is a header of a unified data block:
+
+    +---------------------------------------------------------------+
+    |                             token                             |
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                         encrypted data                        +
+    |                                                               |
+    +---------------------------------------------------------------+
+
+In Token Protect mode the token is detached from encrypted data:
+
+    +---------------------------------------------------------------+
+    |                                                               |
+    +                         encrypted data                        +
+    |                                                               |
+    +---------------------------------------------------------------+
+
+    +---------------------------------------------------------------+
+    |                             token                             |
+    +---------------------------------------------------------------+
+
+And in Context Imprint mode there is no authentication token at all:
+
+    +---------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


ilammy:
(In the good news: I have run out of branches with dusty commits in them so it will be a while until the encryption and decryption parts are documented, exacerbated by the loss of access to the RFCs which have some descriptions already.)

ilammy:
Also, while I've been preparing examples, I have noted a peculiar thing. Look the Soter algorithm descriptor used by Secure Cell in Seal mode: `0x40010100`. That is, AES-GCM-256, with Soter KDF, **and PKCS#7 padding**. Initially I assumed it to be `0x40000100`, with no padding, as PKCS#7 is not really applicable to GCM. But that's the case. I was surprised to learn that Secure Cell uses the `0x40010000` mask for its GCM modes. While GCM uses *a sort of* padding, it's not PKCS#7 one.

I wonder, if we should note it somewhere as yet another hysterical raisin. And it would be nice if Ye Olde Ones could shed some light on it, assuming they do remember some conscious decision around it.

vixentael:
> That is, AES-GCM-256, with Soter KDF, and PKCS#7 padding

_Tangled crypto_

vixentael (approved):
Amazing job!
Even if some details are wrong, still, this represents the best of our knowledge and bias at this time.

themis/spec/secure-cell.md
```diff
@@ -383,3 +383,186 @@ This value is stored as `40 0d 03 00` in little-endian encoding.
...
+
+All the encoded values match the expectations.
+
+### Example: passphrases
+
+With the following inputs (all encoded in ASCII):
+
+| Input        | Value |
+| ------------ | ----- |
+| passphrase   | `au6aimoa8Pee8wahxi4Aique6eaxai2a` |
```
iamnotacake, line 465: `passphrase` here looks the same as `encryption key` in example above, is it intentional?

Also, what does ASCII encoding exactly mean? I guess every char corresponds to one byte in `encryption key`, 32 bytes in total, just like AES-256 expects. And here it is like 32-ASCII-char-string password. A little bit confusing imo.

ilammy: > passphrase here looks the same as encryption key in example above, is it intentional?

Yeah. I have been intentionally lazy to change it :) I guess I'll switch it for something more passphrase-looking, now that you've mentioned it.

> Also, what does ASCII encoding exactly mean?

Well, exactly what it says. You see *text* in the spec but cryptography works with bytes. There are various ways of converting them. In particular, the key might look like it's base32-encoded or something. For avoidance of doubt, I wanted to point out that those should be treated as ASCII strings. That is:

<table>
<tr>
<th>Text string</th><th>Encoded bytes</th>
</tr>
<tr>
<td>
<code>encrypted message</code>
</td>
<td>
<pre>
65 6e 63 72 79 70 74 65 64 20 6d 65 73 73 61 67
65
</pre>
</td>
</tr>
<tr>
<td>
<code>au6aimoa8Pee8wahxi4Aique6eaxai2a</code>
</td>
<td>
<pre>
61 75 36 61 69 6d 6f 61 38 50 65 65 38 77 61 68
78 69 34 41 69 71 75 65 36 65 61 78 61 69 32 61
</pre>
</td>
</tr>
</table>

ilammy: Or, to put in more practical way, these are the strings that you should give to command-line utilities to reproduce the results.

vixentael (approved):

themis/spec/secure-cell.md
```diff
@@ -220,3 +220,166 @@ AES-GCM algorithm has a limit on how much data can be safely encrypted using the
...
+| Encryption algorithm       | Algorithm ID |
+| -------------------------- | ------------ |
+| AES-256-GCM **(default)**  | `0x40010100` |
+| AES-196-GCM                | `0x400100C0` |
+| AES-128-GCM (_deprecated_) | `0x40010080` |
+
+Refer to the [algorithm descriptor overview](../common/#soter-symmetric-algorithm-descriptor) for details.
+
+The **IV data** and **auth tag** fields are theoretically flexible,
+but with current algorithm choice IV is always 12 bytes long
```
storojs72, line 315: Probably, we should explain why IV is 12 bytes length. It is a recommended parameter (in context of GCM mode) for interoperability, efficiency and security (in some sense). We can reference NIST SP-800 38D document for curious readers

Explanation: 
1) https://crypto.stackexchange.com/questions/41601/aes-gcm-recommended-iv-size-why-12-bytes;
2) https://crypto.stackexchange.com/questions/59396/generate-j-0-for-gcm-cipher-when-operatornameleniv-neq-96-bits

themis/spec/secure-cell.md
storojs72: @iamnotacake , nice catch!

storojs72 (approved):

themis/spec/secure-cell.md
ilammy: > Probably, we should explain why IV is 12 bytes length.

Makes sense. Will do so for PBKDF2 salt length too, I remember referencing some other recommendation for it.

vixentael:
Merge 'em all!

PR #32, opened 2020-12-23
Update crypto export guidelines

vixentael:
Add section about Google Play.

P.S. I'd like to rename page URL from `apple-crypto-regulations/` to `us-crypto-export-compliance/` and setup redirect, however it requires to change file names and setup redirect in hugo configs, so I plan to handle this later.

### DIFF
--- a/themis/regulations/apple-crypto-regulations.md
+++ b/themis/regulations/apple-crypto-regulations.md
@@ -5,6 +5,20 @@ title: US crypto export regulations
 
 # US crypto export regulations for Themis
 
+When you distribute apps via platforms like App Store or Google Play,
+the binaries are typically hosted on servers located within US, owned by US companies.
+United States laws treat this activity as export, which is heavily regulated for cryptography.
+
+However, typical Themis use cases fall under “open source” exceptions.
+This makes compliance much easier if your app is open source as well.
+
+{{ note warning }}
+**Note:**
+If your app is not open source or is not distributed free of charge,
+we strongly recommend that you seek professional legal advice.
+{{ /note }}
+
+
 ## Submitting apps to App Store
 
 If your application uses Themis and you want to submit it to the Apple App Store,
--- a/themis/regulations/apple-crypto-regulations.md
+++ b/themis/regulations/apple-crypto-regulations.md
@@ -1,9 +1,9 @@
 ---
 weight: 1
-title:  App Store requirements
+title: App Store requirements
 ---
 
-# Apple encryption export regulations for Themis
+# US crypto export regulations for Themis when submitting to App Store
 
 If your application uses Themis and you want to submit it to the Apple App Store,
 you are required to do the following:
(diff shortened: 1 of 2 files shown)


storojs72 (approved):
Probably we could rename this doc to be more general - since Google Play is related to Android, while App Store - to iOS. What about `cryptographic regulations for mobile applications`?

themis/regulations/apple-crypto-regulations.md
```diff
@@ -1,9 +1,9 @@
 ---
 weight: 1
-title:  App Store requirements
+title: App Store requirements
```
ilammy, line 3: Since this page is going to be about US cryptography export regulations, I think it should have an appropriate title.

```suggestion
title: US crypto export regulations
```

themis/regulations/apple-crypto-regulations.md
```diff
@@ -1,9 +1,9 @@
...
 ---
 
-# Apple encryption export regulations for Themis
+# US crypto export regulations for Themis when submitting to App Store
```
ilammy, line 6: Similarly, I believe there should be only one `<h1>` header on the page which serves as page title. Others should be structured under it. Plus, multiline headers don't look that great.

```suggestion
# US crypto export regulations for Themis

## Submitting apps to App Store
```

themis/regulations/apple-crypto-regulations.md
```diff
@@ -28,7 +28,18 @@ Read more in the official [Apple export compliance overview](https://help.apple.
 and [BIS guidelines](https://bis.doc.gov/index.php/policy-guidance/encryption/1-encryption-items-not-subject-to-the-ear)
 (you might need to use a VPN to access it).
 
-#### Submitting an annual self-classification report to Apple
+
+# US crypto export regulations for Themis when submitting to Google Play
```
ilammy, line 32: Flattening the structure:

```suggestion
## Submitting apps to Google Play
```

themis/regulations/apple-crypto-regulations.md
```diff
@@ -28,7 +28,18 @@ Read more in the official [Apple export compliance overview](https://help.apple.
...
+
+If your application uses Themis and you want to submit it to the Google Play,
+you are required to do the following:
+
+1. Send an annual (year-end) self-classification report to the US government
+    to comply with the [encryption export regulations](https://developer.apple.com/documentation/security/complying_with_encryption_export_regulations).
```
ilammy, line 38: I don't feel it's right to link to Apple docs in the Google Play section. Since Google does not seem to provide more details on the matter other than the Google Play guidelines linked below, I think we can leave only that link.

```suggestion
 1. Send an annual (year-end) self-classification report to the US government
    to comply with the encryption export regulations.
```

Also, I have an OCD for indentation alignment.

ilammy (approved):
@vixentael, I'm good with the spirit of the change, but as usual I have some nitpicking to do 😉

Also, I think that it won't hurt to add a short blurb right under the page title, explaining *why* do I as Themis user need to care for US export laws when submitting apps to App Store and Google Play.

```
When you distribute apps via platforms like App Store or Google Play,
the binaries are typically hosted on servers located within US, owned by US companies.
United States laws treat this activity as export, which is heavily regulated for cryptography.

However, typical Themis use cases fall under “open source” exceptions.
This makes compliance much easier if your app is open source as well.

{{ note warning }}
**Note:**
If your app is not open source or is not distributed free of charge,
we strongly recommend that you seek professional legal advice.
{{ /note }}
```

And also, I think that the parent “Regulations” page need to mention Google Play too. Currently is talks only about App Store.

@storojs72,

> What about cryptographic regulations for mobile applications?

App Store hosts apps for (desktop) macOS too, and those need to be declared as well. You need to do this for *everything*. Like, Linux repositories which serve OpenSSL binaries from servers located within US also fall under these regulations. See [Debian](https://www.debian.org/legal/cryptoinmain), for example.

vixentael:
Thank you both for suggestions, added them! 

> What about cryptographic regulations for mobile applications

US crypto export laws are about any software and hardware that uses crypto and falls under 'export'. In reality, these are quite old laws that are not well suited for the world of modern software. The reason why we have section about App Store / Google Play and not about Debian/Ubuntu distributions is because we receive more questions from mobile developers that use Themis (and we're not legal advisers).

ilammy:
> In reality, these are quite old laws that are not well suited for the world of modern software.

Well, at least now Themis is not considered a munition 😂

PR #34, opened 2021-01-23
Split "spec/common" sections into their own pages

ilammy:
Make a dedicated page for each common concept instead of piling things into the `/themis/spec/common` page.

### Why do this

There will be more and probably more sections added to this page and I'd like to keep it manageable.

For one, I'd like the anchors to have some meaningful names, instead of `#example-1` referring to Soter algorithm descriptor example which will change if you dare to move the section.

At the same time, it's tiresome to ensure that each section has unique name that fits in one line. Sure, you can name it “Example: Soter algorithm descriptor” or “Soter algorithm descriptor example”, but this removes a very useful structure: each page having an “Example” section.

So... Since with the new doc server we're moving away from the “monopage concept”, let's do this to the common section of the spec.

### On URL compatibility

While updating the page URLs I could have added backup compatibility anchors to the `common/` index page, but since it's so small the users wouldn't have an issue locating the next link they should follow.

It would be cool to do such redirects automatically, but we still don't have the machinery for anchors.


vixentael (approved):

vixentael:
Splitting makes a lot of sense :)

PR #35, opened 2021-01-23
Reference implementation of Soter container

ilammy:
Okay, here's my stupid idea. In addition to specs with prose, provide users with a reference implementation of whatever is described. Some people are better at understanding concepts through code. Plus, this provides an "executable specification" and tests. I have found this invaluable when writing Secure Cell spec: that caught quite a few mistakes and misconceptions that I had about its implementation.

(History-savvy people here could have found parallels between this and the concept of "literate programming". Indeed, if only Knuth's ideas caught up, this would have been a perfect opportunity for that approach. Unfortunately, it's not really popular to have good tooling.)

The reference implementation is written in Go. Aside from an obvious reasons having to do with Cossack Labs being a Go shop, this choice has a number of other benefits:

  - very good standard library, with a lot of cryptographic primitives available which will come very handy later
  - readable to most programmers who are familiar with C syntax
  - almost immediately executable if you copy-paste the sample code, thanks to "push to GitHub to publish a package" approach
  - reference implementation can actually reuse its own code for different cryptosystems

Unfortunately, you can't run this on Go Playground without much hassle, but that would have been amazing.

Now, note that "reference implementation" is different from the "official GoThemis source code". The overriding objective here is to illustrate key points in Themis algorithms. Readability and correctness are the main focus of the reference implementation. Performance, usability, good error reporting, feature completeness – all of this can be sacrificed for the sake of brevity and clarity of the code.

Themis Core's C code is anything but easy to understand cryptography. Since even its maintainers make mistakes when deciphering code paths leading to arcane OpenSSL invocations, any cryptography audit would be pretty expensive for it. Hopefully, this reference implementation can serve as a better way to understand the cryptography behind Themis.

Okay, so... Here's an implementation of Soter container for starters. Let's link to it for the implementation details and replace the snippet with something shorter (which still invokes the same code). The reader should be able to copy the snippet, run it with `go run`, and see the output which corresponds to whatever is there in the spec prose.

What do you all think?

Dependencies:
- [X] #34

### DIFF
--- a/themis/spec/common/soter-container_test.go
+++ b/themis/spec/common/soter-container_test.go
@@ -1,0 +1,31 @@
+package common
+
+import (
+	"encoding/base64"
+	"testing"
+)
+
+func TestExampleContainer(t *testing.T) {
+	containerB64 := "VUVDMgAAAC1s1W74A6Sx9yhDygNh4YEb0LShLZrEgTosYF2yRVG4pHGoaa6N"
+	containerBytes, _ := base64.StdEncoding.DecodeString(containerB64)
+
+	container, remaining := ParseSoterContainer(containerBytes)
+	if container == nil {
+		t.Fatal("not a Soter container")
+	}
+	if len(remaining) != 0 {
+		t.Error("unexpected trailing bytes")
+	}
+
+	if container.Tag() != "UEC2" {
+		t.Error("invalid tag")
+		t.Log("actual:  ", container.Tag())
+		t.Log("expected:", "UEC2")
+	}
+
+	if len(container.Payload) != 33 {
+		t.Error("invalid payload length")
+		t.Log("actual:  ", len(container.Payload))
+		t.Log("expected:", 33)
+	}
+}
--- a/themis/spec/common/soter-container_test.go
+++ b/themis/spec/common/soter-container_test.go
@@ -6,6 +6,12 @@ import (
 )
 
 func TestExampleContainer(t *testing.T) {
+	// This is an EC public key for Themis, the one dissected in the specification.
+	// You can use Themis command-line tools to generate a new keypair:
+	//
+	//     go run docs/examples/go/secure_keygen.go | tail -n +2
+	//
+	// See https://docs.cossacklabs.com/themis/debugging/cli-utilities/
 	containerB64 := "VUVDMgAAAC1s1W74A6Sx9yhDygNh4YEb0LShLZrEgTosYF2yRVG4pHGoaa6N"
 	containerBytes, _ := base64.StdEncoding.DecodeString(containerB64)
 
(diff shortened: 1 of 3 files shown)


vixentael:
The idea looks really attractive.

So, the main benefit is that future-us would read short&easy-to-understand Go reference implementations instead of long&ugly C code, right? So, documentation purposes mostly?

themis/spec/common/soter-container_test.go
```diff
@@ -0,0 +1,31 @@
+package common
+
+import (
+	"encoding/base64"
+	"testing"
+)
+
+func TestExampleContainer(t *testing.T) {
```
Lagovas, line 8: imho, better to test container implementation using C themis library. Generate some container by source of trust library and try to parse it here to proof that it is correct implementation and it is compatible with C themis. Because without this check it is one more container that lives separately from everyone.

ilammy:
@vixentael,

> So, the main benefit is that future-us would read short&easy-to-understand Go reference implementations instead of long&ugly C code, right? So, documentation purposes mostly?

Yeah, that's the goal. *Ideally*, the production implementation is readable enough to be usable as a reference, but it gets hairy pretty quickly. Current GoThemis can't be used because it's just a wrapper. Even if it weren't, there will be many more “extra” code to maintain backwards compatibility, additional error checking, good error reporting, optional features, etc. In a reference implementation used for documentation purposes we can throw away all that fluff to focus on the cryptography primitives.

themis/spec/common/soter-container_test.go
ilammy: > imho, better to test container implementation using C themis library

That's exactly what is being done here: the container is a public key that I have generated with actual Themis.

I believe that a proper interoperability test suite can be implemented later, but it should stay in a separate package from this one. That will allow to keep *this* implementation with exactly zero external dependencies, immediately available to be `go run` or `go test`. A test suite for interoperability with actual GoThemis will require Themis Core to be installed in the system, at least. I don't want this to be a prerequisite to step through a test you're interested in when reading the spec.

However, this is really nice idea to implement later: ensure that both the reference and actual implementation behave the same, at least on the happy path. Thanks for pointing it out!

Lagovas: oh, I didn't know that base64 value is container generated by C themis. can you add comment about it? or describe in comment how to generate the same. For example using gothemis in golang and what to copy\paste for that.

ilammy: > can you add comment about it?

Well, sure, here you go.

I did not generate *this* particular value in the described way, but command-line tools can be most definitely used to generate a new key and wrap it in base64 for compactness.

vixentael (approved):

Lagovas (approved):

PR #40, opened 2021-03-14
Soter KDF description & reference implementation

ilammy:
Describe Soter KDF algorithm which is used primarily in Secure Cell. Note that it's similar in construction to ZRTP KDF but not quite the same. Also note the differences, as well as the idiosyncratic "implicit key derivation" process.

Provide a reference implementation, some tests and test vectors for it.

### DIFF
--- a/themis/spec/common/soter-kdf.go
+++ b/themis/spec/common/soter-kdf.go
@@ -1,0 +1,49 @@
+package common
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+)
+
+// SoterKDF derives a new key of requested length from an input key.
+// Label indicates key derivation purpose.
+// Derived key is also bound to provided context data (if any).
+// Output key length must be between 1 and 32 bytes.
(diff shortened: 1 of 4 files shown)


themis/spec/common/soter-kdf.go
```diff
@@ -0,0 +1,49 @@
...
+	mac.Write([]byte{0x00})
+	for _, context := range context {
+		mac.Write(context)
+	}
+	result := mac.Sum(nil)
+	// Return the result truncated to desired output key size.
+	return result[:outputBytes]
+}
+
+func xor(out, in []byte) []byte {
```
Lagovas, line 40: not obvious that we need separate `xor` function because golang doesn't have `min` function  ) because golang has ^ operator

ilammy: Uh... Okay, here you go, I've rewritten it with `min()` and `^=`.

Lagovas (approved):

vixentael (commented):
If @storojs72 doesn't have any comment, I suggest merging

PR #49, opened 2021-05-16
Reference implementation of Soter algorithm ID

ilammy:
Throw it in for completeness. This one is easy to implement and the spec is very clear so I won't reference it directly. However, other systems working with Soter algorithm IDs will benefit from this part for sure.

### DIFF
--- a/themis/spec/common/soter-alg.go
+++ b/themis/spec/common/soter-alg.go
@@ -1,0 +1,74 @@
+package common
+
+// AlgorithmID encodes symmetric algorithm parameters for Soter.
+type AlgorithmID uint32
+
+const (
+	algorithmMask   = 0xF0000000
+	algorithmOffset = 28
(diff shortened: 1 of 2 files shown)


vixentael (approved):

vixentael:
@Lagovas take a look pls

Lagovas:
looks good, but will be better if we use gothemis in tests and validate actual values generated by library instead of hardcoded examples. for example, create ciphertext with default mode and check that all these masks and values synchronized with the current version of library.

ilammy:
> but will be better if we use gothemis in tests and validate actual values generated by library instead of hardcoded examples

We'll get there once Secure Cell algorithms are implemented. Then it will be possible to get GoThemis encrypt something and use the reference implementation to inspect parsed structure of Secure Cells.

PR #59, opened 2021-07-13
Acra General Configuration docs

iamnotacake:
### Configuring & maintaining > General configuration
* Necessary preparations (described how to generate master key, what else do we need here?)
* AcraTranslator (described CLI flags and how to generate different keys used by acra-translator, what else?)
* AcraServer (CLI flags)

CLI flags are currently grouped in three groups (`jaeger`, `logging`, `vault`) + TLS flags on another page (guess I'll move them here) + the rest of flags outside of any group.


acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,308 @@
...
+
+* `--auth_keys=<filename>`
+
+  Path to basic auth credentials.
+  To add user, use: `./acra-authmanager --set --user <user> --pwd <pwd>`.
+  Default is `configs/auth.keys`.
+
+* `--client_id=<id>`
+
+  Expected client ID of AcraConnector in mode without encryption.
```
Lagovas, line 51: it's client_id used by AcraServer for transparent encryption in environment without acra-connector. acra-server uses this value as like it took from acra-connector.

acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,308 @@
...
+* `--postgresql_enable={true|false}`
+
+  Handle Postgresql connections.
+  Default is `false`.
+
+* `--redis_db_keys=`
+
+  Number of Redis database for keys.
+  Default is `0`.
+  <!-- `acra-translator -help` says default is `-1` but in `cmd/redis.go` I see `redisDefaultDB = 0` -->
```
Lagovas, line 175: hmm... it is interesting, why --help shows `-1` as default... I looked how these flags were declared and yes, used constant with value `0`...

Lagovas (approved):

acra/configuring-maintaining/general-configuration/necessary_prep.md
```diff
@@ -0,0 +1,38 @@
...
+$ cat /tmp/master_key
+{"encryption":"cgVd5kHock5dy2bBGv6dHmHCgi2Piq3+y43lE7tTgh0=","signature":"fNNgVVBvjaIuL7Ea5lNcs4Ccz8SqLF4bXGhJvLGVL0k="}
+```
+
+This key should then be base64-encoded and passed to Acra services in `ACRA_MASTER_KEY` environment variable.
+
+Like this: `ACRA_MASTER_KEY="$(cat /tmp/master_key | base64)" acra-server ...`
+
+There are other kinds of keys, for example ones specific for AcraTranslator.
+You will find more about these keys on their page.
```
storojs72, line 28: Should we add some useful links here?

storojs72 (approved):
LGTM!

Minor: I suppose that `Jaeger` / `Logging` / `Vault` flags are identical for both AcraServer and AcraTranslator - may be we can have them specified in a single separate place like for TLS?

storojs72:
I would a bit clarify the minor point from this comment: https://github.com/cossacklabs/product-docs/pull/59#pullrequestreview-706629054

I would prefer to keep information about flags for AcraServer/AcraTranslator in subsequent pages (even if some flags are similar), but probably have some additional pages about unified approaches that we have chosen for TLS/logging/metrics of both Acra services

acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,419 @@
...
+
+* `--db_port=<port>`
+
+  Database port for AcraServer -> database connections.
+  Default is `5432`.
+
+* `-ds`
+
+  Turn on HTTP debug server.
+  The server will be listening at `127.0.0.1:6060`.
```
ilammy, line 65: I'd move this to the “logging” section as it's more fitting to be there, as a debug tool.

acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,419 @@
...
+  Export trace data to log.
+  Default is `false`.
+
+* `-v`
+
+  Log to stderr all `INFO`, `WARNING` and `ERROR` logs.
+
+* `-d`
+
+  Log everything to stderr.
```
ilammy, line 387: I'd put short options first, followed by long options. These are short on purpose, you might need them more often.

acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,419 @@
...
+  Logging format.
+
+  * `plaintext` — (default) pretty human readable key/value format  
+    `time="2021-07-12T14:02:12+03:00" level=info msg="Starting service acra-translator [pid=475995]" version=0.85.0`
+
+  * `json` — one JSON object per line, easy to parse by most log collectors  
+    `{"level":"info","msg":"Starting service acra-translator [pid=476077]","product":"acra-translator","timestamp":"2021-07-12T14:02:50+03:00","unixTime":"1626087770.004","version":"0.85.0"}`
+
+  * `CEF` — Common Event Format  
+    `CEF:0|cossacklabs|acra-translator|0.85.0|100|Starting service acra-translator [pid\=476133]|1|unixTime=1626087782.510`
```
ilammy, line 374: Fenced blocks would render more nicely:

```markdown
  * `plaintext` — (default) pretty human readable key/value format

    ```
    time="2021-07-12T14:02:12+03:00" level=info msg="Starting service acra-translator [pid=475995]" version=0.85.0
    ```
```

Also, it would be nice to stop this trailing whitespace thing. A lot of editors would strip them by default and won't show them. Use `<br>` if you need a line break.

acra/configuring-maintaining/general-configuration/acra_server.md
```diff
@@ -0,0 +1,419 @@
...
+## Command line flags
+
+* `--acracensor_config_file=<filename>`
+
+  Path to AcraCensor configuration file.
+
+* `--acraconnector_tls_transport_enable={true|false}`
+
+  Enable TLS to encrypt transport between AcraConnector and AcraServer.
+  Default is `false` which means "use SecureSession instead".
```
ilammy, line 18: This kitchensink of flags can be subdivided further:

* Network (`--db_host`, `--incoming_connection`)
* Keystore (`--auth_keys`, `--keys_dir`, `--keystore`)
* Caching
* MySQL configuration
* PostgreSQL configuration
* AcraConnector stuff
* AcraServer features (`--acrastruct_injectedcell_enable`, `--zonemode_enable`)
* whatever else you can see

Ditto for other binaries.

acra/configuring-maintaining/general-configuration/acra_translator.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Keys
+
+There are few different kinds of keys for different purposes.
+When generating them, few things should be kept in mind:
+* `ACRA_MASTER_KEY` should be set containing master key of same keystore version
+* Output directory should have no permissions for group and others
+
+`client1` in examples is the identifier of client that can be either sent in
+RPC requests or derived from the client TLS certificate depending on configuration.
```
ilammy, line 264: IMO, this should go on some other page, not mixed with listing of command-line options.

acra/configuring-maintaining/general-configuration/necessary_prep.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+# Necessary preparations (keys, ports, etc.)
+
+## Master key
+
+Most Acra services require master key in order to work and perform cryptographic operations.
+It is used, for example, to encrypt other keys in their keyrings.
+These keys are then used in more specific tasks
+(encrypt data in database, hash data for searchable encryption).
```
ilammy, line 14: I'd say these are completely separate topics:

- generating and managing your master key
- managing Acra keystore (which is related)
- what ports you should open for Acra to function

And they should be covered on separate pages, I believe.

acra/configuring-maintaining/general-configuration/acra_server.md
Lagovas: but it does not provide any logs. it is just an HTTP server for `go pprof` tool. not sure that it is obvious to search for such options in `logging` section... but we don't have a lot of options for debugging purposes to move them to a separate section. imho, we can leave it among other options for now.

acra/configuring-maintaining/general-configuration/necessary_prep.md
```diff
@@ -0,0 +1,43 @@
...
+This key should then be base64-encoded and passed to Acra services in `ACRA_MASTER_KEY` environment variable.
+
+Like this: `ACRA_MASTER_KEY="$(cat /tmp/master_key | base64)" acra-server ...`
+
+{{< hint warning >}}
+Make sure the master key is stored in a safe place.
+It's more important than any other key.
+{{< /hint >}}
+
+There are other kinds of keys, for example [ones specific for AcraTranslator](../acra_translator/#keys).
```
Lagovas, line 32: better to use absolute links from the root using `{{< ref "/acra/configuring-maintaining/general-configuration/...." >}}` tag because it will safe us from future problems like copy-paste/moving content with relative links to another place. because after that link will be related to another path and will be broken

iamnotacake: That doesn't work, maybe a bug, maybe something else. I tried
- `{{< ref "/acra/configuring-maintaining/general-configuration/acra_translator/#keys" >}}`
- `{{< ref "/acra/configuring-maintaining/general-configuration/acra_translator" >}}`

and `hugo` gives error `page not found`.

On the other hand:
- `{{< ref "/acra/configuring-maintaining/general-configuration" >}}` works
- `/acra/configuring-maintaining/general-configuration/acra_translator/#keys` in the web browser opens that exact page I'm pointing to
- so decided to change `../acra_translator/#keys`
  to `/acra/configuring-maintaining/general-configuration/acra_translator/#keys` w/o this hugo magic

acra/configuring-maintaining/general-configuration/acra_server.md
iamnotacake: `Network`, `Keystore` — done.
`Caching` — what except `--keystore_cache_size` (that already belongs to `Keystore`) is about caching?
`MySQL`, `PostgreSQL` — done, but they only got one flag to enable mysql, one to enable pgsql, two deprecated flags for pgsql, do we need sections for these?

BTW do we need to move Redis flags to `Keystore`? Since it is used as key DB. Or is it more about caching?

PR #58, opened 2021-07-12
Installation section, AcraStruct, Trying Docker

Lagovas:
Migrated `Getting started` + `Quickstart` sections from old doc to new as `Installation` with several pages that are standalone and was easy to migrate: `Trying Acra with docker` and `Digital Ocean marketplace`, plus migrated AcraStruct description to AcraBlocks that was mentioned in Installation.

I didn't migrate `Write an app with AcraWriter` from `Quickstart`. I think it should be migrated into other sections or rewritten.

PR #60, opened 2021-07-20
Migrate security design 

Lagovas:
Here I migrated page from [source docserver](https://docs.cossacklabs.com/pages/security-design/)
These changes were forked from previous PR #58 , so now it will be draft until PR will be merged. Here is only one new file and you can review only added [commit](https://github.com/cossacklabs/product-docs/commit/ca13ddc105863938f0411c9e75de396d5d5d8501).


storojs72 (approved):
Reviewed mentioned in description commit. LGTM

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
 title: Security design
 bookCollapseSection: true
 ---
+
+## Security design
+
+Acra is built to provide selective encryption only to the records that require such protection.
+AcraWriter provides the app developer with the means to encrypt these records via wrapping anything in a function that outputs [AcraStruct]({{< ref "acra/acra-in-depth/data-structures/#understanding-acrastruct" >}}), a crypto container decryptable by AcraServer, which is then stored in a database.
+
+Any secret contained within an app (i.e. passwords in config files) can be leaked. There are many reasons why a developer might not want to use the user password as a secret - not all systems have to be end-to-end encrypted, even though we'd like it to be that way. An example of a good trust model is to use asymmetric cryptography to "send" this data (for the encryption of which we only have a public key) to some remote party. A leakage of this public key will still be insufficient for decrypting anything from the database.
```
ilammy, line 11: > **i.e.** passwords in config files

This should be “e.g.”

i.e. = *id est*, “that is”, used to explain or rephrase something
e.g. = *exempli gratia*, “for example”, used to introduce examples

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+### Threat model
+
+We expect that anything, apart from AcraServer (which must remain uncompromised), can be compromised and that any piece of data can leak outside the system.
+
+Our goal is to provide the two following guarantees:
+
+**Guarantee 1**: If the attacker compromises the system through collecting all the data from any (or all) components other than AcraServer, and AcraServer remains uncompromised, this data will not be sufficient to decrypt protected entries.
+
+More specifically, AcraServer's key storage is the 'crown jewel' of the whole security scheme and should be protected with the typical server security measures.
```
ilammy, line 21: > with the typical server security measures.

This reads like you can run everything else with no security thought at all (= not ”typical measures”), `root/root` and public open ports, and still end up totally fine. I don't think Acra works that way (yet?)

Instead, I think you should make extra effort to secure the AcraServer and its keys.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+- Data leaks caused by mistakes in deploy, configuration settings, backup thefts, etc.
+- SQL injections.
+- Denial of service.
+- Vulnerabilities in the database protocol.
+- Weak audit or an absence thereof.
+- Operational system vulnerabilities.
+- Unsafe handling of cryptographic keys.
+
+### Conditions of secure work
+Acra can perform its protective functions properly and protect from the security threats 1, 2, 3 if the following security assumptions are met:
```
ilammy, line 98: > Conditions of secure work

“Preconditions” or “Basic assumptions” would be better, I think. The current wording reads somehow unnatural.

> the security threats 1, 2, 3

I guess this refers to the most common threats listed above, but the list is not numbered. It will be better to just name the threats here once again, it won't hurt.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+- The PKI infrastructure is trusted;
+- AcraServer is trusted;
+- The client is less trusted than the server.
+
+### Possible consequences of compromisation
+Let’s consider all the possible consequences of any of separate component being broken (broken as in “fully compromised” when the adversary fully overtakes the work of the component and gains full access to its memory).
```
ilammy, line 104: > Possible consequences of **compromisation**

I don't think “compromisation” is a widely used word. “Data breach” or “security breach” would be something more common and easier to grasp.

> **all** the possible consequences

Does this cover zombie infestation and the DBAs losing their jobs? If no, I guess it should be just “possible consequences”.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+When a *Database* is broken into, the worst-case scenario is DoS or COA. Thus, the stability of the system, in this case, is reduced to the stability of the symmetric encryption algorithm (AES-GCM-256).
+When the *Client* gets broken, the worst-case scenario is that the adversary can get the data belonging to this client, which is stored in the database.
+And finally, if AcraServer gets broken, the adversary can fully compromise the system.
```
ilammy, line 108: It would be better to define what “broken” means in this context. Data dump on the public internet? Leaked keys? Both? Another heartbleed found? Fundamental flaw in AES discovered? Attacker gaining control over the process? The server put out of service by DoS? The datacenter being on fire?

Also, I think COA needs to be expanded since it's not a common abbreviation. At least I had no clue what it means, [the Wikipedia list](https://en.wikipedia.org/wiki/COA) suggests that ciphertext-only attack might be it. I think this page would benefit from being more accessible to non-cryptographers.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+It is worth mentioning that in absence of PKI, the communication channel between the Client and AcraServer is also vulnerable. In this case, the resistance ability of the system comes down to the secureness of the SSL/TLS or Themis’ Secure Session protocols. In all the other communication channels the data is encrypted so, in the worst case (when SSL/TLS is not used) the secureness of the system comes down to the secureness of the symmetric encryption algorithm (AES-GCM-256).
```
ilammy, line 110: > secureness

“security” would be a more common word, as in “security of X”.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+### Additional reading
+
+We recommend that you also check out the following articles to gain a better understanding of the security notions in this article: <br>
+https://www.zdnet.com/article/the-top-ten-most-common-database-security-vulnerabilities/ <br>   
+https://en.wikipedia.org/wiki/Vulnerability_database <br>           
+https://www.bcs.org/content/ConWebDoc/8852. <br>
```
ilammy, line 117: Unordered lists work just fine and look better.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+We strongly recommend following the best security practices and using the public key infrastructure (PKI) to provide authentication between the components of Acra. This document contains some useful information about the PKI and links to related resources.
+
+### The Public Key Infrastructure
+
+The central component of any PKI is the Certificate Authority (CA). The CA acts as an arbitrator or a third party, whilst the trust relations are established between the two other sides. A very simplified scheme of one-side authentication of `User` by a `Relying Party` is described below.
+
+
+
+![](/files/wiki/pki-overview-for-acra.png)
+
```
ilammy, line 131: Same thing as elsewhere about placing files under `files/wiki` and about `{{ ref }}` usage.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+(_Image source: [LdapWiki](http://ldapwiki.com/wiki/Public%20Key%20Infrastructure)._)
+
+Each entity of the PKI possesses their own pair of cryptographic keys:
+- Private key that is always stored secretly, while
+- Public key, along with the corresponding digital certificate, is available for other entities.
+
+The digital certificate is signed by the CA and contains the information about an entity (organisation, department, alias/name, email, etc.) and the value of its public key. The digital certificate also indicates that the entity possesses a corresponding private key.
+
+> Note: The CA issues certificate for its own public key, which is why it is considered to be the main point of trust: both the User and the Relying Party trust the CA, while the User and the Relying Party do not trust each other.
+
```
ilammy, line 141: Same thing as elsewhere about using `{{< hint info >}}` blocks.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+The main point that should be kept in mind is that if you use a free-for-all communication infrastructure (i.e. the Internet), some primary security layer (PSL) that involves a PKI (i.e. Virtual Private Network) has to be deployed (i.e. Virtual Private Network (VPN)). This PSL will provide a strong authentication between components (i.e. the database, [AcraServer]({{< ref "acra/configuring-maintaining/general-configuration/acra-server.md#-INVALID" >}}) in Acra’s context) inside the organisational infrastructure.
```
ilammy, line 142: > (i.e. Virtual Private Network)

Just once is enough.

(And same thing as elsewhere about “e.g.” vs “i.e.”)

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+We recommend checking out the following pages and resources to find out more about the PKIs.
+
+1) https://pki-tutorial.readthedocs.io/en/latest/
+2) https://openvpn.net/
+3) https://www.ejbca.org/
+4) https://sourceforge.net/projects/xca/
+5) https://github.com/OpenVPN/easy-rsa
+6) https://en.wikipedia.org/wiki/Public_key_infrastructure
+7) https://en.wikipedia.org/wiki/Virtual_private_network.
```
ilammy, line 166: Nit: This could be an unordered list.

Also, a stray dot makes the last link broken.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+- **Files** — many technologies of secrets’ deployment  use the transfer of secrets in the file format to the app - both for infrastructure management systems and container orchestration systems. This method is usually implemented through a regular creation of files containing secrets (at the moment deployment or by daemons of the key distribution systems), or through mounting from the host system inside a container or virtual file system mounting (Swarm, Kubernetes). The pros and cons and the same as in the variant with the environment variables, with some specifics depending on the exact implementation.
+
+- **URI / REST API** — another rather common way based on cloud services like AWS, Azure, DO, etc. The main disadvantage of this approach is its price.
+
+- **Libs** — usually implemented using the supplied libraries. They are offered by AWS, Azure, etc., to enable the work with their KMS clouds. The standard de-facto approach here is Vault integrated with numerous applications.
```
ilammy, line 152: > `- **Libs**`

I'd say KMS instead. It's not *just any* library that would work. This is more about various connector libraries to access some key management service.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,165 @@
...
+
+Acra currently supports secret delivery of keys via environment variables since this is the simplest and the most reliable way.
```
ilammy, line 154: Isn't it more complicated that that? Some keys can be read from environment variables, some keys can be read from files, accessed over the network, or fetched from KMS by Acra EE.

acra/acra-in-depth/security-design/_index.md
Lagovas: There is the main focus on keys in context of security. Same is true for PKI and root keys. Your environment that uses PKI should be protected too. But the core of the security schema is the root keys used for PKI that open access for whole environment, not for one specific node/app/host.

acra/acra-in-depth/security-design/_index.md
Lagovas: > “Preconditions” or “Basic assumptions” would be better, I think. The current wording reads somehow unnatural.

what difference between "conditionals" and "preconditionals" in this context? May be better use "requirements" instead of "conditions"... @storojs72 , what do you think?

> I guess this refers to the most common threats listed above, but the list is not numbered. It will be better to just name the threats here once again, it won't hurt.

we have numbered guarantees above and it refers to them, and they haven't been named. How you can name "compromisation systems except AcraServer" ?)

acra/acra-in-depth/security-design/_index.md
Lagovas: > I don't think “compromisation” is a widely used word. “Data breach” or “security breach” would be something more common and easier to grasp.

I tried to google it and IMHO, it's a common word in data/app security context, used in news, articles, etc.
https://www.securityweek.com/search/google/compromise?query=compromise&cx=016540353864684098383%3A6mcx-eenlzi&cof=FORID%3A11&sitesearch=&safe=off
@storojs72 , what do you think? IMHO "data breach" is more often used in general hype news when something was hacked without knowing what was done. But compromisation is a more specific action.

> Does this cover zombie infestation and the DBAs losing their jobs? If no, I guess it should be just “possible consequences”.

there we describe what we mean in parentheses. I don't know why you think about zombies in docs for data security tool ) and why someone will think about same too.

acra/acra-in-depth/security-design/_index.md
Lagovas: > It would be better to define what “broken” means in this context

actually, all of these things and we mention the worst of them on our opinion that related to data security.

> Also, I think COA needs to be expanded since it's not a common abbreviation.

agree, changed to `COA (ciphertext-only attack)`

acra/acra-in-depth/security-design/_index.md
Lagovas: agree. cannot google such phrases in articles/wiki.... changed to "security of"

acra/acra-in-depth/security-design/_index.md
Lagovas: you can use KMS (for example) from infrastructure/deployment scripts (URI/REST API example), and you can use it from applications via "libs". Or you can derive keys from some master key that stored in ENV/file using libraries (not implemented by themselves)
I understand it like that...

any other opinions? @storojs72 , @vixentael ?

acra/acra-in-depth/security-design/_index.md
Lagovas: I'm not sure, but for DevOps it may be the simplest way. Just because Env variables doesn't have permissions like files, doesn't need any accounts in KMS or configuration for network. Just set and use) Same for developers.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+- Private key that is always stored secretly, while
+- Public key, along with the corresponding digital certificate, is available for other entities.
+
+The digital certificate is signed by the CA and contains the information about an entity (organisation, department, alias/name, email, etc.) and the value of its public key. The digital certificate also indicates that the entity possesses a corresponding private key.
+
+{{< hint info >}}
+Note: The CA issues certificate for its own public key, which is why it is considered to be the main point of trust: both the User and the Relying Party trust the CA, while the User and the Relying Party do not trust each other.
+{{< /hint >}}
+
+The main point that should be kept in mind is that if you use a free-for-all communication infrastructure (e.g. the Internet), some primary security layer (PSL) that involves a PKI (e.g. Virtual Private Network (VPN)) has to be deployed (e.g. VPN). This PSL will provide a strong authentication between components (e.g. the database, [AcraServer]({{< ref "acra/configuring-maintaining/general-configuration/acra-server.md#-INVALID" >}}) in Acra’s context) inside the organisational infrastructure.
```
vixentael, line 147:
```suggestion
The main point that should be kept in mind is that if you use a free-for-all communication infrastructure (e.g. the Internet), some primary security layer (PSL) that involves a PKI has to be deployed (e.g. VPN). This PSL will provide a strong authentication between components (e.g. the database, [AcraServer]({{< ref "acra/configuring-maintaining/general-configuration/acra-server.md#-INVALID" >}}) in Acra’s context) inside the organisational infrastructure.
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+
+- **Environment variables** — classic and reasonably secure way that is simple, flexible, and can be implemented with system tools. The main known risks of this method are: the possibility to bleed into logs if the deployment was untidy and the possible granting of access to the non-target apps in case of insufficiently fine-tunes access rights differentiation.
+
+- **Files** — many technologies of secrets’ deployment  use the transfer of secrets in the file format to the app - both for infrastructure management systems and container orchestration systems. This method is usually implemented through a regular creation of files containing secrets (at the moment deployment or by daemons of the key distribution systems), or through mounting from the host system inside a container or virtual file system mounting (Swarm, Kubernetes). The pros and cons and the same as in the variant with the environment variables, with some specifics depending on the exact implementation.
+
+- **URI / REST API** — another rather common way based on cloud services like AWS, Azure, DO, etc. The main disadvantage of this approach is its price.
```
vixentael, line 155:
```suggestion
- **URI / REST API** — another rather common way based on cloud services like AWS, Azure, DO, etc. The main disadvantage of this approach is its price. Cloud KMS also often use REST API to delivery secrets.
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+
+- **Libs** — usually implemented using the supplied libraries. They are offered by AWS, Azure, etc., to enable the work with their KMS clouds. The standard de-facto approach here is Vault integrated with numerous applications.
```
vixentael, line 157:
```suggestion
- **KMS** — usually implemented using the supplied libraries of KMS providers. They are offered by AWS, Azure, etc., to enable the work with their KMS clouds. The standard de-facto approach here is HashiCorp Vault integrated with numerous applications.
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+
+Acra currently supports secret delivery of keys via environment variables since this is the simplest and the most reliable way.
```
vixentael, line 159:
```suggestion
Acra supports several ways of delivering keys – direct connection to HashiCorp Vault; delivery via environment variables.
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+- Operational system vulnerabilities.
+- Unsafe handling of cryptographic keys.
+
+### Conditions of secure work
+Acra can perform its protective functions properly and protect from the security threats 1, 2, 3 if the following security assumptions are met:
+- The PKI infrastructure is trusted;
+- AcraServer is trusted;
+- The client is less trusted than the server.
+
+### Possible consequences of compromisation
```
vixentael, line 105:
```suggestion
### Possible consequences of security incident
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+Let’s consider all the possible consequences of any of separate component being broken (broken as in “fully compromised” when the adversary fully overtakes the work of the component and gains full access to its memory).
```
vixentael, line 106:
```suggestion
Let’s consider possible consequences of any of separate component being broken (broken as in “fully compromised” when the adversary fully overtakes the work of the component and gains full access to its memory).
```

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+
+When a *Database* is broken into, the worst-case scenario is DoS or COA (ciphertext-only attack). Thus, the stability of the system, in this case, is reduced to the stability of the symmetric encryption algorithm (AES-GCM-256).
+When the *Client* gets broken, the worst-case scenario is that the adversary can get the data belonging to this client, which is stored in the database.
+And finally, if AcraServer gets broken, the adversary can fully compromise the system.
+
+It is worth mentioning that in absence of PKI, the communication channel between the Client and AcraServer is also vulnerable. In this case, the resistance ability of the system comes down to the security of the SSL/TLS or Themis’ Secure Session protocols. In all the other communication channels the data is encrypted so, in the worst case (when SSL/TLS is not used) the security of the system comes down to the security of the symmetric encryption algorithm (AES-GCM-256).
```
vixentael, line 112:
```suggestion
It is worth mentioning that in absence of PKI, the communication channel between the Client and AcraServer is also vulnerable. In this case, the resistance ability of the system comes down to the security of the TLS or Themis’ Secure Session protocols. In all the other communication channels the data is encrypted so, in the worst case (when TLS is not used) the security of the system comes down to the security of the symmetric encryption algorithm (AES-GCM-256).
```

acra/acra-in-depth/security-design/_index.md
vixentael: "compromisation" – this is a very rarely used word in security context. Say breach, incident, compromise, not compromisation.

acra/acra-in-depth/security-design/_index.md
```diff
@@ -2,3 +2,170 @@
...
+- Threats related to excess / abuse of access privileges.
+- Data leaks caused by mistakes in deploy, configuration settings, backup thefts, etc.
+- SQL injections.
+- Denial of service.
+- Vulnerabilities in the database protocol.
+- Weak audit or an absence thereof.
+- Operational system vulnerabilities.
+- Unsafe handling of cryptographic keys.
+
+### Conditions of secure work
```
vixentael, line 99:
```suggestion
### Security assumptions
```

vixentael (changes requested):

Lagovas:
@vixentael , I committed your suggestions manually. missed that github allow commit them as a batch now

vixentael (approved):

acra/acra-in-depth/security-design/_index.md
storojs72: I think, @ilammy 's point targets particularly wording "typical security measures" which would be better to rephrase like "special security measures"

acra/acra-in-depth/security-design/_index.md
storojs72: I would agree that term "data breach" or "hacking" is usually used in non-formal sources (hype news as well), although at recent time "breach" term starts appearing in applied RnD papers like [this](https://arxiv.org/pdf/1703.02014.pdf) where they use "system breach" and "data breach" as similar meanings.

In more formal sources (especially crypto-related) I usually meet "adversarial breaking" term (for example [here](https://www.cs.virginia.edu/dwu4/courses/sp20/static/reductions.pdf)). It denotes to situation when the security of system / cryptographic scheme is formalized in a form of some mathematical problem and adversary under some circumstances is able to "break security".

Having that, I would not mind against "breach", but IMO I would use "Analyzing of possible consequences of Acra breaking" for naming of this section

acra/acra-in-depth/security-design/_index.md
storojs72: Sorry for late review iteration. I agree with "Security assumptions" renaming of this small section. However I would expand it a bit by clarifying what "trusted" means for each item in list

acra/acra-in-depth/security-design/_index.md
storojs72: I'm not an expert, so would agree with "KMS". I assume that difference between KMS and URI / REST API is that with KMS user have more control over it's keys (it can deploy own in-premise server and install Vault there for keeping keys) while with URI / REST API - key storing is delegating to cloud provider. Is it?

PR #63, opened 2021-07-29
CLI flags for acra-connector

Lagovas:
Here are described CLI flags for acra-connector. Used descriptions for common flags from #59 and style of describing default values. Please take a look at `--mode` flag. I'm not sure that it is the best way to note that some flags work only with a specific mode.

### DIFF
--- a/acra/configuring-maintaining/general-configuration/acra_connector.md
+++ b/acra/configuring-maintaining/general-configuration/acra_connector.md
@@ -17,16 +17,29 @@ weight: 4
 
   Provided Client ID identifies Secure Session transport keys for handshake with AcraServer/AcraTranslator and will be sent to AcraServer/AcraTranslator as identifier of encryption keys
 
-* `--mode`
-
-  Expected mode of connection. 
-  * `AcraServer` - (default) mode switch AcraConnector to work with `AcraServer`. Only in this mode works next flags: `incoming_connection_port`/`incoming_connection_api_port`/`acraserver_connection_host`/`acraserver_connection_string`/`http_api_enable`/`acraserver_api_connection_string`/`acraserver_api_connection_port`/`acraserver_securesession_id`
-  * `AcraTranslator` - mode switch AcraConnector to work with `AcraTranslator`. Only in this mode works next flags: `acratranslator_connection_host`/`acratranslator_connection_string`/`acratranslator_connection_port`/`acratranslator_securesession_id`
-
 * `--user_check_disable`
 
   Disable checking that connections from app running from another user
 
+* `--mode`
+
+  Expected mode of connection.
+  * `AcraServer` - (default) mode switch AcraConnector to work with `AcraServer`. 
+  * `AcraTranslator` - mode switch AcraConnector to work with `AcraTranslator`. 
+
+Following table describes flags that work only in appropriate mode:
+
+| AcraServer | AcraTranslator |
+| --- | --- |
+| `acraserver_connection_host` | `acratranslator_connection_host` |
+| `acraserver_connection_string` | `acratranslator_connection_string` |
+| `acraserver_connection_port` | `acratranslator_connection_port` |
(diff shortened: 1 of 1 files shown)


acra/configuring-maintaining/general-configuration/acra_connector.md
```diff
@@ -0,0 +1,337 @@
...
+  `<host:port>` used to connect to Redis.
+  Default is empty (don't connect).
+
+* `--redis_password=<password>`
+
+  Password to Redis database.
+
+### TLS
+
+> Note: TLS related flags works only together with `AcraServer` mode. `AcraTranslator` mode supports only `SecureSession` transport encryption or direct TLS connections from application
```
storojs72, line 230: "flags works" -> "flags work"

acra/configuring-maintaining/general-configuration/acra_connector.md
```diff
@@ -0,0 +1,337 @@
...
+
+* `--client_id=<id>`
+
+  Provided Client ID identifies Secure Session transport keys for handshake with AcraServer/AcraTranslator and will be sent to AcraServer/AcraTranslator as identifier of encryption keys
+
+* `--mode`
+
+  Expected mode of connection. 
+  * `AcraServer` - (default) mode switch AcraConnector to work with `AcraServer`. Only in this mode works next flags: `incoming_connection_port`/`incoming_connection_api_port`/`acraserver_connection_host`/`acraserver_connection_string`/`http_api_enable`/`acraserver_api_connection_string`/`acraserver_api_connection_port`/`acraserver_securesession_id`
+  * `AcraTranslator` - mode switch AcraConnector to work with `AcraTranslator`. Only in this mode works next flags: `acratranslator_connection_host`/`acratranslator_connection_string`/`acratranslator_connection_port`/`acratranslator_securesession_id`
```
storojs72, line 24: Here we can have a primitive table:

| AcraServer | AcraTranslator |
| --- | --- |
| incoming_connection_api_port | acratranslator_connection_host |
| acraserver_connection_host | acratranslator_connection_string |      
| acraserver_connection_string | acratranslator_connection_port |      
| http_api_enable | acratranslator_securesession_id |      

But I'm not sure if it will be rendered correctly for hugo

storojs72 (commented):
The problem with `--mode` flag (and in general with big number of flags in Acra services) is that user will definitely confuse what combination of flags to use. AFAIK, to tackle with such problems subcommands can be introduced. Having `mode` as a subcommand we can enforce which flags are available for AcraServer / AcraTranslator and I guess it should be possible to specify common flags shared by both subcommands like TLS, Vault, etc. However it will require some redesigning of our Acra services' main functions. Previously we had some discussions around using [cobra](https://github.com/spf13/cobra) for subcommands introducing - and for some reasons we refused AFAIR.

ilammy:
> The problem with --mode flag (and in general with big number of flags in Acra services) is that user will definitely confuse what combination of flags to use. AFAIK, to tackle with such problems subcommands can be introduced. Having mode as a subcommand we can enforce which flags are available for AcraServer / AcraTranslator and I guess it should be possible to specify common flags shared by both subcommands like TLS, Vault, etc.

How about just separate `acra-server-connector` and `acra-translator-connector` binaries? That way it could be more obvious which matches what.

acra/configuring-maintaining/general-configuration/acra_connector.md
Lagovas: what about moving "mode" flag to the end of section with table? and that is how it looks:
image

storojs72:
> How about just separate `acra-server-connector` and `acra-translator-connector` binaries? That way it could be more obvious which matches what

That would also require some engineering. But may be less amount of work in comparing with subcommands. I think we will not go this direction right now

storojs72 (approved):
LGTM

PR #66, opened 2021-08-10
acra-keymaker cli description

Lagovas:


acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+* `--tls_cert=<path>`
+  
+  Path to TLS certificate to use as client_id identifier. Works only when passed empty ClientID: `--client_id=""`.
+
+* `--tls_identifier_extractor_type={distinguished_name|serial_number}`
+  
+  Decide which field of TLS certificate to use as ClientID.
+
+  * `distinguished_name` — (default) certificate Distinguished Name (DN).
+  * `serial_number` — certificate serial number.
```
ilammy, line 25: “(default) certificate Distinguished Name” reads like something you would choose during certificate issuance. It would be nice to tack “(default)” at the end but you already put “(DN)” there, so I guess you can just note the default later.

```suggestion
  * `distinguished_name` — certificate Distinguished Name (DN)
  * `serial_number` — certificate serial number
  
By default certificate Distinguished Name is used as client ID.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--keystore=<v1|v2>`
+  
+  Set keystore format. If `--keys_output_dir` points to existing keystore destination then `acra-keymaker` automatically detects version of keystore.
+  If destination storage is empty and hasn't any key then this argument is required. Read more about [keystore versions]({{< ref "/acra/keys/versions/" >}})
```
ilammy, line 30: It would be better to make it more explicit that this needs to be used only for new keystores:

```suggestion
* `--keystore=<v1|v2>`
  
  Set keystore format.
  Read more about [keystore versions]({{< ref "/acra/keys/versions/" >}}).
  
  This option is required for newly created keystores.
  If `--keys_output_dir` points to existing keystore then `acra-keymaker` automatically detects the version of the keystore.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--keys_output_dir=<folder>`
+  
+  Folder where will be saved keys (you can change destination for asymmetric public keys with keystore V1 (V2 store at same destination)  with parameter `--keys_public_output_dir`).  
+  Default is `.acrakeys`.
+
+* `--keys_public_output_dir=<folder>`
+  
+  Folder where will be saved public asymmetric keys. Works only for `--keystore=v1` and do nothing with `--keystore=v2`.
+  Default is `.acrakeys`.
```
ilammy, line 44: You need to pick either “directory” or ”folder” and stick with it. (Also, you can keep things more simple. (Also, there are (too) many (nested) parentheses here, (I think).))

```suggestion
* `--keys_output_dir=<path>`

  Path to keystore directory.

  Default is `.acrakeys`.

* `--keys_public_output_dir=<folder>`

  Path to public key directory for `--keystore=v1` format.
  (Ignored for `--keystore=v2`.)

  Default is `.acrakeys`.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+  
+#### Redis
+
+* `--redis_db_keys=`
+
+  Number of Redis database for keys.
+  Default is `0`.
+  <!-- `acra-server -help` says default is `-1` but in `cmd/redis.go` I see `redisDefaultDB = 0` -->
+  <!-- this var is also used as default value for the flag, where's the truth? -->
```
ilammy, line 53: I think a placeholder won't hurt here. Also, “number of Redis database” reads like you mistyped “databases” and meant something related to quantity of those databases. The technical term is “database number” so we need to feature it somehow...

```suggestion
* `--redis_db_keys=<number>`

  Redis database number to use.
  Default is `0`.
  <!-- `acra-server -help` says default is `-1` but in `cmd/redis.go` I see `redisDefaultDB = 0` -->
  <!-- this var is also used as default value for the flag, where's the truth? -->
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--redis_host_port=<host:port>`
+
+  `<host:port>` used to connect to Redis.
+  Default is empty (don't connect).
```
ilammy, line 58: You can just call it “address”, IMO.

```suggestion
* `--redis_host_port=<host:port>`

  Address of Redis database to use as keystore.
  If not specified, Redis is not used.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+* `--redis_password=<password>`
+
+  Password to Redis database.
+
+
+### Configuration files
+
+* `--config_file=<filename>`
+
+  Path to YAML configuration file.
```
ilammy, line 69: Consistency. In most cases it's paths, not just file names. The placeholder suggests what type of data you pass there.

```suggestion
* `--config_file=<path>`

  Path to YAML configuration file.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--dump_config`
+
+  Dump configuration to `configs/acra-translator.yaml`.
+
+* `--generate_markdown_args_table`
+
+  Generate markdown file with text description of all flags.
+  Output file is `configs/markdown_acra-translator.md`.
+  Works in pair with `--dump_config`.
```
ilammy, line 79: Uh... I think you meant `acra-keymaker` here?
```suggestion
* `--dump_config`

  Dump configuration to `configs/acra-keymaker.yaml`.

* `--generate_markdown_args_table`

  Generate markdown file with text description of all flags.
  Output file is `configs/markdown_acra-keymaker.md`.
  Works in pair with `--dump_config`.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Vault
+
+* `--vault_connection_api_string=<url>`
+
+  Connection string (like `http://x.x.x.x:yyyy`) for loading `ACRA_MASTER_KEY` from HashiCorp Vault.
+  Default is empty (`ACRA_MASTER_KEY` environment variable is expected).
```
ilammy, line 86: If you insist on HashiCorp Vault™ branding everywhere, the header should reflect that.

Also, I think it's enough to mention just once how Vault is used by Acra.

Also, using HTTP is bad practice. I don't think Vault recommends that (but IIRC, it *is* possible to disable it).

```suggestion
### HashiCorp Vault

`acra-keymaker` can read `ACRA_MASTER_KEY` from HashiCorp Vault instead of environment variable.

* `--vault_connection_api_string=<url>`

  Connection string (like `https://example.com:8200`) for connecting to HashiCorp Vault.
  If not specified, `ACRA_MASTER_KEY` environment variable will be used.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--vault_secrets_path=<path>`
+
+  KV Secret Path for reading `ACRA_MASTER_KEY` from HashiCorp Vault.
+  Default is `secret/`.
```
ilammy, line 91: You need to distinguish this path from filesystem paths somehow.

```suggestion
* `--vault_secrets_path=<kv-path>`

  Path to KV Secrets directory in Vault used to store `ACRA_MASTER_KEY`.
  Default is `secret/`.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--vault_tls_client_cert=<filename>`
+
+  Path to client TLS certificate for reading `ACRA_MASTER_KEY` from HashiCorp Vault.
+  Default is empty (don't send client certificate).
+
+* `--vault_tls_client_key=<filename>`
+
+  Path to private key of the client TLS certificate for reading `ACRA_MASTER_KEY` from HashiCorp Vault.
+  Default is empty (don't send client certificate).
```
ilammy, line 106: Same thing as elsewhere with “paths” vs “filenames”. Also, “default is empty” is not really appropriate here because “empty value” by itself means a special action.

```suggestion
* `--vault_tls_ca_path=<path>`

  Path to CA certificate bundle to use for HashiCorp Vault certificate validation.

  If not specified, use root certificates configured in system.

* `--vault_tls_client_cert=<path>`

  Path to client TLS certificate used to connect to HashiCorp Vault.

  If not specified, don't send client certificate.

* `--vault_tls_client_key=<path>`

  Path to the private key of the client TLS certificate used to connect to HashiCorp Vault.

  If not specified, don't send client certificate.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--vault_tls_transport_enable={true|false}`
+
+  Use TLS to encrypt transport with HashiCorp Vault.
+  Default is `false`.
```
ilammy, line 111: > Default is `false`

Questionable defaults, IMO.

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+### Common keys
+
+{{< hint warning >}}
+Note: backup all your keys before calling `acra-keymaker` with any flag described below. Some of them rotate and save current keys, 
+some of them may override current keys.
+{{< /hint >}}
+
+* `--generate_master_key`
+  
+  Create new random master key acceptable by all Acra tools and save to file.
```
ilammy, line 122: You need to tell which file will be used. It's also a nice place to remind how this file is supposed to be used.

```suggestion
* `--generate_master_key=<path>`

  Generate a new random master key and write it to the specified file.

  All Acra components will expect this key in `ACRA_MASTER_KEY` environment variable.
```

(Tricky question: how do you tell `acra-keymaker` to put the key into Vault instead?)

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+  Use TLS to encrypt transport with HashiCorp Vault.
+  Default is `false`.
+  
+### Common keys
+
+{{< hint warning >}}
+Note: backup all your keys before calling `acra-keymaker` with any flag described below. Some of them rotate and save current keys, 
+some of them may override current keys.
+{{< /hint >}}
```
ilammy, line 118: I don't think there are any other common keys except for the master key. It would be better to make this whole section about generating keys.

Also, grammar. And the warning can stand out a bit more.

```suggestion
## Generating keys

{{< hint warning >}}
**Note:**
Back up all your keys before calling `acra-keymaker` with any flag described below.
Some keys can be rotated and the previous keys will be retained,
while some keys will be overwritten.
{{< /hint >}}

### Acra master key
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `--generate_master_key`
+  
+  Create new random master key acceptable by all Acra tools and save to file.
+
+{{< hint info >}}
+Note: When you generate any new type of key (or keypair) described below it rotates current key. After that new key becomes default 
+for encryption purposes in [Transparent proxy mode]({{< ref "/acra/configuring-maintaining/general-configuration/acra_server.md#transparent-proxy-mode-INVALID" >}}) 
+and archive previous key that used only for decryption purposes
+{{< /hint >}}
```
ilammy, line 128: Some grammar tweaks.

```suggestion
{{< hint info >}}
**Note:**
When you generate a new key (or keypair), the existing key is rotated.
That is, the new key becomes the default for encryption in [Transparent proxy mode]({{< ref "/acra/configuring-maintaining/general-configuration/acra_server.md#transparent-proxy-mode-INVALID" >}})
while the previous key is archived and used only for decryption.
{{< /hint >}}
```

Also, what happens to transport keys when they are rotated? I believe the old keys are retained as well, but I don't think services use them.

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+  
+  Create new transport keypair for AcraConnector.
+
+* `--generate_acraserver_keys`
+  
+  Create new transport keypair for AcraServer.
+
+* `--generate_acratranslator_keys`
+  
+  Create new transport keypair for AcraTranslator.
```
ilammy, line 142: Insistent terminology.

```suggestion
* `--generate_acraconnector_keys`
  
  Generate a new transport keypair for AcraConnector.

* `--generate_acraserver_keys`
  
  Generate a new transport keypair for AcraServer.

* `--generate_acratranslator_keys`
  
  Generate a new transport keypair for AcraTranslator.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+  
+  Create key symmetric key for secure logging.
+  
+* `--generate_poisonrecord_keys`
+  
+  Create keypair and symmetric key for poison records.
+  
+* `--generate_symmetric_storage_key`
+  
+  Create symmetric key for data encryption/decryption with AcraBlock.
```
ilammy, line 166: More insistent terminology.

```suggestion
* `--generate_acrawriter_keys`
  
  Generate a new storage keypair for data encryption/decryption with AcraStruct.
  
* `--generate_hmac_key`
  
  Generate a new key for HMAC calculation and searchable encryption.
  
* `--generate_log_key`
  
  Generate a new key symmetric key for audit logging.
  
* `--generate_poisonrecord_keys`
  
  Generate a new keypair and symmetric key for poison records.
  
* `--generate_symmetric_storage_key`
  
  Generate a new symmetric key for data encryption/decryption with AcraBlock.
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+### AcraServer & AcraAuthManager keys
+
+* `--generate_acrawebconfig_keys`
+  
+  Create symmetric key for encryption AcraWebconfig's basic auth credentials.
+  {{< hint warning >}}
+  Note: this key is not rotatable and this flag replaces current with new one.
+  {{< /hint >}}
+
+
```
ilammy, line 178: Even more insistent terminology (and grammar).

```suggestion
* `--generate_acrawebconfig_keys`
  
  Generate a new symmetric key for encrypting AcraWebconfig's basic authentication credentials.

  {{< hint warning >}}
  **Note:** This key cannot be rotated. Newly generated key replaces the current one.
  {{< /hint >}}
```

acra/configuring-maintaining/general-configuration/acra_keymaker.md
Lagovas: yes, they archived and not used anymore

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,178 @@
...
+  
+  Use provided client ID as identifier to generate keys or keypairs.
+
+* `--tls_cert=<path>`
+  
+  Path to TLS certificate to use as client_id identifier. Works only when passed empty ClientID: `--client_id=""`.
+
+* `--tls_identifier_extractor_type={distinguished_name|serial_number}`
+  
+  Decide which field of TLS certificate to use as ClientID.
```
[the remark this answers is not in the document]
Lagovas: agree, will use ClientID because remember that we use it in other pages

acra/configuring-maintaining/general-configuration/acra_keymaker.md
Lagovas: hm, really. will look on this why we did so. maybe were some reasons. If not, will change and probably deprecate and deny usage without TLS

acra/configuring-maintaining/general-configuration/acra_keymaker.md
Lagovas: nice question. created task for it) because it's good feature to allow generate and save master key directly to vault.

acra/configuring-maintaining/general-configuration/acra_keymaker.md
Lagovas: Hm... In our --help it written as `generate` and `create` and I left only `Create`. But now look at this one more time and yes, better to use `Generate` because our commands use verb `generate_...` )

acra/configuring-maintaining/general-configuration/acra_keymaker.md
```diff
@@ -0,0 +1,196 @@
...
+
+  Use TLS to encrypt transport with HashiCorp Vault.
+  Default is `false`.
+  
+## Generating keys
+
+{{< hint warning >}}
+**Note:**
+Back up all your keys before calling `acra-keymaker` with any flag described below.
+Some keys can be rotated and the previous keys will be retained,
```
storojs72, line 131: May be refer here to key rotation section for more details?

storojs72 (approved):

PR #68, opened 2021-08-17
Add optimization docs

iamnotacake:
What was done:
* Horizontal scaling (just link to demo)
* Comparison of AcraStructs and AcraBlocks
* Recommend reusing connection to DB
* Disable debug logs
* Mention indexes required for good searchable encryption performance
* Using LRU caches
* TLS configuration and its impact on performance
* Obtaining metrics
* Tracing (was not part of this PR initially)

Not for this PR:
* Vertical scaling (I guess better to wait for T1956 to be completed and take data from there)
* Enterprise kickstart


acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,27 @@
...
+weight: 3
+---
+
+# AcraStructs vs AcraBlocks
+
+There are two different "crypto containers" AcraServer can use to store encrypted value in.
+Then, these containers are stored in database and decrypted back by AcraServer when needed.
+
+History
+* `AcraStruct` is our "old" one, though we're not going to remove it any time soon.
```
Lagovas, line 12: It's not an old one. It is one of two. AcraStructs main crypto container for client-side encryption for now. Applications can encrypt with ephemeral asymmetric keys on app side with AcraStructs, but cannot do this with AcraBlocks because of symmetric key encryption. But they can configure AcraServer to re-encrypt AcraStructs (that generated on app side) on AcraServer side. Because AcraStructs allow to be sure that data are encrypted when it is transmitted between app and AcraServer.

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,27 @@
...
+* `AcraBlock` on the other hand is new and proposes some extra features.
+
+Encryption with a key
+* `AcraStruct` uses key provided from somewhere outside (by Acra) to encrypt the value.
```
Lagovas, line 16: uses storage public key that related to private key known only for AcraServer. Main difference between AcraStruct and AcraBlock that first uses asymmetric key encryption that are slow due to multiplications of big numbers (elliptic curve encryption) and symmetric key encryption that just XOR arrays using optimized instructions in CPU.

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,27 @@
...
+* `AcraBlock` uses random key to encrypt the value, then encrypts this key with another key, this time provided from outside.
+  That first key is called DEK (data encryption key) and the second one is called KEK (key encryption key).
+
+Key rotation (when you need re-encrypt data with a new key)
+* With `AcraStruct`, the value has to be decrypted completely, then encrypted with a new key.
```
Lagovas, line 21: actually AcraStructs uses KEK and DEK too. But for KEK used asymmetric encryption that are slow. And you need to load private key, extract public key from AcraStruct, compute ECDHE to get "shared key" that used as KEK. And after that decrypt DEK.
When for AcraBlock AcraServer uses symmetric key and just decrypts KEK that encrypted and stored in AcraBlock. And decryption uses symmetric encryption that are much faster than asymmetric.

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,27 @@
...
+* With `AcraBlock`, only DEK should be re-encrypted, leaving the encrypted value the same.
+  And this may have a big impact on performance when rotating keys for a large amount of encrypted data.
+
+Storage overhead
+* `AcraStruct` uses a bit less extra bytes compared to `AcraBlock`.
```
Lagovas, line 26: hm.... i'm not sure that AcraStructs uses less than AcraBlock.
```
package main

import (
	"fmt"
	"github.com/cossacklabs/acra/acrablock"
	"github.com/cossacklabs/acra/acrastruct"
	"github.com/cossacklabs/themis/gothemis/keys"
)
import _ "github.com/lib/pq"

func main() {
	testData := []byte(`a`)
	ecKeys, err := keys.New(keys.TypeEC)
	if err != nil {
		panic(err)
	}
	acraStruct, err := acrastruct.CreateAcrastruct(testData, ecKeys.Public, nil)
	if err != nil {
		panic(err)
	}
	acraBlock, err := acrablock.CreateAcraBlock(testData, []byte(`some key`), nil)
	if err != nil {
		panic(err)
	}
	fmt.Println(len(acraStruct), len(acraBlock))
}
// >> 190 139
```

AcraStruct: 190
AcraBlock 139

acra/configuring-maintaining/optimisations/connection_reuse.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Connection reuse
+
+Most of libraries/frameworks for work with SQL databases have a dedicated object for the connection with database.
+The one you get after soccessful connetion to database with used credentials.
+It is recommended to **reuse** such object as much as possible.
+
+Reusing connections will result in lower latencies (no need to wait for connection before every SQL request).
+
+It will also reduce the load on AcraServer and the database behind since they won't have to handle new connections so frequently.
```
Lagovas, line 14: we can say that it will reduce amount of memory that AcraServer uses to handle connections. Because it works similar to any database that allocate temporary buffers for each new connection. AcraServer does it too.

acra/configuring-maintaining/optimisations/db_indexes.md
```diff
@@ -0,0 +1,23 @@
...
+## Searchable encryption
+
+When some column is configured to be "encrypted and searchable", AcraServer will store its value as `HMAC(value) + encrypt(value)`.
+This `HMAC(value)` is the hash of value that is used during search of encrypted value (knowing the `value` exactly).
+For this to be efficient, there should be a special index in database that indexes only **first 32 bytes** of encrypted-searchable column.
+`32` because of the hash being used (SHA256).
+
+In SQL the index would look something like
+```sql
+CREATE INDEX searchable_email ON users (SUBSTR(email, 0, 32))
```
Lagovas, line 21: `substr(email, 1, 33)` works for PostgreSQL because it start indexing from 1 (actually 0 means same as 1, but 32 byte has 33 index). And I don't remember how it works in MySQL, better to check it.

Plus we can note that no reason to create indexes for not searchable AcraStructs/AcraBlocks because they are random arrays and indexes will slow down writing and will not increase performance for reading.

acra/configuring-maintaining/optimisations/connection_reuse.md
iamnotacake: Changed, is new version better?

acra/configuring-maintaining/optimisations/db_indexes.md
iamnotacake: Changed, is new version better?

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
iamnotacake: Changed, is new version better?
(talking about storage overhead comparison, other issues in this file will be fixed in next commit)

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
iamnotacake: Turned out I don't know their internal layout so well. Gonna update based on description from "Data structures" page.

acra/configuring-maintaining/optimisations/_index.md
```diff
@@ -2,3 +2,30 @@
...
 ---
+
+[Vertical scaling]({{< ref "acra/configuring-maintaining/optimisations/vertical_scaling.md" >}})
+describes how performance will grow with more CPU cores / memory added
+
+[Horizontal scaling]({{< ref "acra/configuring-maintaining/optimisations/horizontal_scaling.md" >}})
+describes how performance will grow with instances being added
+
+[AcraStructs vs AcraBlocks]({{< ref "acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md" >}})
+describes difference between AcraStructs and AcraBlocks (two possible "containers" for encrypted database columns)
```
Lagovas, line 13: not only database columns. AcraTranslator doesn't operates with columns, only with user data. Maybe just `encrypted data`?

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+
+Acra services provide some metrics for Jaeger when configured to do so. This includes
+[AcraConnector]({{< ref "acra/configuring-maintaining/general-configuration/acra_connector/_index.md#jaeger" >}}),
+[AcraServer]({{< ref "acra/configuring-maintaining/general-configuration/acra_server/_index.md#jaeger" >}}),
+[AcraTranslator]({{< ref "acra/configuring-maintaining/general-configuration/acra_translator/_index.md#jaeger" >}}).
+
+# Metrics
+
+<!-- decryptor/postgresql/pg_decryptor.go and decryptor/mysql/response_proxy.go -->
+* `ProxyClientConnection` checks every client request using AcraCensor, if request is allowed, sends it to the database (PostgreSQL or MySQL)
```
Lagovas, line 16: it's time of processing incomming connection from client to database through AcraServer. This trace starts after accepting TCP connection + SecureSession handshake if used AcraConnector and ends after closing connection from client, database or AcraServer side.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Obtaining metrics
+weight: 9
+---
+
+# Obtaining metrics
+
+Acra services provide some metrics for Jaeger when configured to do so. This includes
```
Lagovas, line 8: it is not a metrics but traces. Jaeger works with traces. Metrics we export for prometheus.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+[AcraConnector]({{< ref "acra/configuring-maintaining/general-configuration/acra_connector/_index.md#jaeger" >}}),
+[AcraServer]({{< ref "acra/configuring-maintaining/general-configuration/acra_server/_index.md#jaeger" >}}),
+[AcraTranslator]({{< ref "acra/configuring-maintaining/general-configuration/acra_translator/_index.md#jaeger" >}}).
+
+# Metrics
+
+<!-- decryptor/postgresql/pg_decryptor.go and decryptor/mysql/response_proxy.go -->
+* `ProxyClientConnection` checks every client request using AcraCensor, if request is allowed, sends it to the database (PostgreSQL or MySQL)
+  * `ProxyClientConnectionLoop` corresponds to one exact query being checked and processed
```
Lagovas, line 17: This trace represents one iteration of processing packet from client's db driver to database and ends on finishing processing. Driver establishes connection with AcraServer/DB and it may be idle some time until it sends new request serialized as several packets in db specific protocol. This trace tracks one db packet processing.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+    * `censor` corresponds to the stage when query is being checked by AcraCensor
+
+<!-- decryptor/postgresql/pg_decryptor.go -->
+* `PgDecryptStream` (PostgreSQL-specific) processes data (rows) from database
```
Lagovas, line 21: it's same as `ProxyClientConnection` but from database side to client through AcraServer. After client's connection AcraServer connects to database and then start proxying. But this trace start tracking only after first database response on client request in PostgreSQL and after establishing connection with db in MySQL mode. Because in PostgreSQL client should send first packet, in MySQL variant database sends first packet.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+  * `PgDecryptStreamLoop` (PostgreSQL-specific) corresponds to one query response being processed
+
+<!-- decryptor/mysql/response_proxy.go -->
+* `ProxyDatabaseConnection` MySQL-specific alternative to PgDecryptStream
```
Lagovas, line 25: right. we forgot rename `PgDecryptStreamLoop` to general name `ProxyDatabaseConnectionLoop` )

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+  * `ProxyDatabaseConnectionLoop` MySQL-specific alternative to PgDecryptStreamLoop
+
+<!-- network/trace.go -->
+* `HTTPApiConnection` TODO
```
Lagovas, line 29: actually this trace not works correctly now. It started tracing but not ends correctly... We should check it and probably fix, because I'm not sure that it exported. So better to mark it "TODO" and comment.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+
+<!-- network/proxy.go -->
+* `ProxyWithTracing` TODO
```
Lagovas, line 32: This trace tracks lifetime of connections in AcraConnector. It starts track on incoming client's connection and new connection from AcraConnector to AcraServer. And finished on close. So it show time of life proxied connections on AcraConnector side. But it start tracking after TLS/Secure Session handshake

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+
+<!-- cmd/acra-connector/acra-connector.go -->
+* `handleConnection` TODO
```
Lagovas, line 35: same as ProxyWithTracing but starts tracking before TLS/SecureSession handshakes. So it wraps whole connection processing and initializing on AcraConnector.

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+  * `WrapClient` TODO
```
Lagovas, line 36: This trace track only handshakes. Start before and ends after handshake, so it measure time of TLS/SecureSession handshakes on AcraConnector

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+
+<!-- cmd/acra-server/common/client_commands_session.go -->
+* `HandleSession` TODO
```
Lagovas, line 39: tracks time of processing HTTP API to AcraServer (not AcraTranslator!). Our HTTP handler doesn't support keep-alive and long connections so this trace measure time of processing all kind of requests (zone generation/fetching current part of config/set new config/LRU cache cleanup/fetching auth data for WebConfig)

acra/configuring-maintaining/optimisations/obtaining_metrics.md
```diff
@@ -0,0 +1,44 @@
...
+
+<!-- cmd/acra-server/common/listener.go -->
+* `WrapServer` TODO
```
Lagovas, line 42: track time of TLS/SecureSession handshakes on AcraServer side

acra/configuring-maintaining/optimisations/tls_configuration.md
```diff
@@ -0,0 +1,37 @@
+---
+title: TLS configuration
+weight: 5
+---
+
+# TLS configuration
+
+Depending on TLS configuration being used, the may be some performance difference:
```
storojs72, line 8:
```suggestion
There are some performance differences depending on TLS configuration being used:
```

acra/configuring-maintaining/optimisations/tls_configuration.md
```diff
@@ -0,0 +1,37 @@
...
+  The key used during certificate generation and in server configuration should be EC one as well.
+
+* Certificate revocation checks
+
+  You may decide to use [OCSP](https://en.wikipedia.org/wiki/Online_Certificate_Status_Protocol)
+  and/or [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list) to increase security of your system.
+  The drawback of using these things is increased TLS handshake time:
+
+  * for every OCSP request, the will be a network request (plain HTTP over TCP)
+  * for every CRL, there will be download of CRL file (again, HTTP)
```
storojs72, line 31:
```suggestion
for every CRL, CRL file will be downloaded (again, HTTP)
```

acra/configuring-maintaining/optimisations/tls_configuration.md
```diff
@@ -0,0 +1,37 @@
...
+  Compared to RSA, it has smaller keys (less data to transmit during handshake), better performance, and at least the same level of security.
+  The key used during certificate generation and in server configuration should be EC one as well.
+
+* Certificate revocation checks
+
+  You may decide to use [OCSP](https://en.wikipedia.org/wiki/Online_Certificate_Status_Protocol)
+  and/or [CRL](https://en.wikipedia.org/wiki/Certificate_revocation_list) to increase security of your system.
+  The drawback of using these things is increased TLS handshake time:
+
+  * for every OCSP request, the will be a network request (plain HTTP over TCP)
```
storojs72, line 30:
```suggestion
for every OCSP request, additional network request will be performed (plain HTTP over TCP)
```

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,44 @@
...
+title: AcraStructs vs AcraBlocks
+weight: 3
+---
+
+# AcraStructs vs AcraBlocks
+
+There are two different "crypto containers" AcraServer can use to store encrypted value in.
+Then, these containers are stored in database and decrypted back by AcraServer when needed.
+
+`AcraStruct` is the main crypto container for the client-side encryption.
```
ilammy, line 11: I'd leave `` `backticks` `` flair for code examples and other verbatim uses. I think using PascalCase is enough to make AcraStruct to stand out as a term.

On a side note, while looking for a way to escape those, here's what I found out about GitHub parser:

### <code><code><code><code><code><code><code><code><code><code><code><code><code><code><code><code>code</code></code></code></code></code></code></code></code></code></code></code></code></code></code></code></code>

By the way, the string above is

    `` `backticks` ``

acra/configuring-maintaining/optimisations/tls_configuration.md
```diff
@@ -0,0 +1,37 @@
...
+  * for every CRL, there will be download of CRL file (again, HTTP)
+
+  Even though CRLs can be cached, at the time when cached item expired, it will be downloaded again, resulting in loger TLS handshake.
```
storojs72, line 33:
```suggestion
Even though CRLs can be cached, at the time when cached item expired, it will be downloaded again, resulting in longer TLS handshake.
```

acra/configuring-maintaining/optimisations/tls_configuration.md
```diff
@@ -0,0 +1,37 @@
...
+
+  It's not a good idea to disable these revocation checks if you need them.
```
storojs72, line 35:
```suggestion 
If OCSP/CRL validations are a bottleneck in performance, you can try to make them faster by:
- providing OCSP/CRL servers in different locations close to clients;
- making sure they are powerful enough to process big amount of requests
```

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,44 @@
...
+Applications can encrypt with ephemeral asymmetric keys on their side with `AcraStruct`s,
+but cannot do this with `AcraBlock`s because of symmetric key encryption.
```
ilammy, line 13: This reads like blaming AcraBlocks for not doing what they were explicitly designed not to do.

```suggestion
AcraStruct is the main crypto container for the client-side encryption.
AcraStructs use ephemeral asymmetric keys to enable applications to encrypt the data but never decrypt it –
the private key stays on the AcraServer which is the only one capable of decryption.
```

And you can add a blurb about AcraBlocks in a separate paragraph after this one. For example,

```markdown
AcraBlocks use only symmetric cryptography to they are generally faster then AcraStructs.
However, this comes at a cost of having to do stricter key management,
since now the application has the decryption keys as well which is a risk.
```

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,44 @@
...
+You can also configure AcraServer to re-encrypt `AcraStruct`s (that were generated on app side)
+on AcraServer side, this way data will definitely be encrypted while transmitted between
+some client app and AcraServer.
```
ilammy, line 16: > this way data will definitely be encrypted while transmitted between some client app and AcraServer.

It's it already encrypted by the transport layer in flight in case of transparent encryption?

I think the point of reencrypting AcraStructs is to make sure that you can't match anything known to the application to the data stored in the database. So an attacker can't submit some AcraStruct, expect it to be stored verbatim, then peek into the database for what row it has been stored at, and then learn something about the system from that.

Another point is preventing usage of compromised keys. AcraServer might accept AcraStructs encrypted with old keys (not necessarily compromised but rotated out), and then reencrypt the data with the current ones.

@Lagovas, @storojs72, what to you think about this part of the paragraph?

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,44 @@
...
+
+Both crypto containers have ability to re-encrypt DEK without touching the encrypted data.
+However, `AcraStruct` will generally be slower because it has to perform more computation
+to decrypt the DEK and encrypt it again with another ephemeral keypair.
+As was mentioned above, this is because of asymmetric cryptography being used (need to compute
+shared ECDH secret, generate ECDH keypair etc).
+
+### Storage overhead (not counting encrypted data length)
+* `AcraStruct` takes about `189` additional bytes
+* `AcraBlock` takes about `138` additional bytes
```
ilammy, line 38: “Takes” usually means in “in total” but here we're talking about overhead, so how about this?

```suggestion
### Storage overhead (in addition to encrypted data length)

* `AcraStruct` adds about `189` extra bytes
* `AcraBlock` adds about `138` extra bytes
```

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
```diff
@@ -0,0 +1,44 @@
...
+
+`AcraStruct` is the main crypto container for the client-side encryption.
+Applications can encrypt with ephemeral asymmetric keys on their side with `AcraStruct`s,
+but cannot do this with `AcraBlock`s because of symmetric key encryption.
+You can also configure AcraServer to re-encrypt `AcraStruct`s (that were generated on app side)
+on AcraServer side, this way data will definitely be encrypted while transmitted between
+some client app and AcraServer.
+
+### Encryption with a key
+
```
ilammy, line 19: I don't really like this header (it's don't say anything to me), but I don't have any real suggestions either. IDK, “Key wrapping”? Something that *actually* relates to the matter discussed in this section. “Encryption with a key” is not it. It's obvious that you need a key for encryption. But the section is not about that, it's about how both containers have a DEK but they deal with KEK differently.

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
storojs72: > You can also configure AcraServer to re-encrypt AcraStructs (that were generated on app side) on AcraServer side, this way data will definitely be encrypted while transmitted between some client app and AcraServer

I agree, it sounds a bit odd. If AcraStruct came to AcraServer from client app, then plaintext from app is definitely encrypted - so there is no need to additionally ensure that data is encrypted. @ilammy's explanation sounds more relevant - with mentioned re-encryption attacker is not able to push specifically crafted AcraStructs to AcraServer and then observe them in the database, since after re-encryption AcraStruct will have different value.

Lagovas: +1, you are right

acra/configuring-maintaining/optimisations/acrastructs_vs_acrablocks.md
iamnotacake: Rewrote this file a lot

Lagovas (approved):
lgtm

PR #70, opened 2021-08-27
Add masking documentation

iamnotacake:
* Short description
* Configuration
* Examples of work

### DIFF
--- a/acra/security-controls/masking/_index.md
+++ b/acra/security-controls/masking/_index.md
@@ -23,32 +23,36 @@ These options are accepted:
 <!-- Config struct lives in encryptor/config/encryptionSettings.go -->
 <!-- Config validation func lives in masking/common/patterns.go -->
 ```yaml
-{
-    # (required, must be non-empty)
-    # Replacement string to use for masking out encrypted data.
-    # If decryption fails, this string will be show instead of the masked
-    # portion of the data.
-    masking: "XXXXXXXXXXXX",
+schemas:
+  - table: table_name
+    encrypted:
+      - column: masked_column_name
 
-    # (required, must be non-negative)
-    # How many bytes of plaintext to leave unencrypted.
-    plaintext_length: 4,
+        # (required, must be non-empty)
+        # Replacement string to use for masking out encrypted data.
+        # If decryption fails, this string will be show instead of the masked
+        # portion of the data.
+        masking: "XXXXXXXXXXXX"
 
-    # (required, allowed values: "left", "right")
-    # Which side of the plaintext to leave unencrypted.
-    # "plaintext_length" bytes will be retained from left or right side of
-    # the data cell.
-    plaintext_side: "right",
+        # (required, must be non-negative)
+        # How many bytes of plaintext to leave unencrypted.
+        plaintext_length: 4
 
-    # (optional, default: "acrastruct", allowed values: "acrastruct", "acrablock")
-    # Which cryptographic container to use for data encryption.
-    crypto_envelope: "acrastruct",
+        # (required, allowed values: "left", "right")
+        # Which side of the plaintext to leave unencrypted.
+        # "plaintext_length" bytes will be retained from left or right side of
+        # the data cell.
+        plaintext_side: "right"
 
-    # (optional, default: false)
-    # If true, data stored in AcraStructs will be transparently reencrypted
-    # into AcraBlocks.
-    reencrypting_to_acrablocks: false,
-}
+        # (optional, default: "acrastruct",
+        # allowed values: "acrastruct", "acrablock")
+        # Which cryptographic container to use for data encryption.
+        crypto_envelope: "acrastruct"
+
+        # (optional, default: false)
+        # If true, data stored in AcraStructs will be transparently reencrypted
+        # into AcraBlocks.
+        reencrypting_to_acrablocks: false
 ```
 <!-- TODO add link to page where colemn encryption settings are described in general, with client_id, zone_id etc -->
 
@@ -56,16 +60,14 @@ These options are accepted:
 
 Here are few examples of configuration and results it would give.
 The actual encrypted part of data will look differently in database, and will take more bytes.
-Examples below only ilustrate which part of plaintext would be encrypted.
+Examples below only illustrate which part of plaintext would be encrypted.
 
 ### Retain first five characters
 
 ```yaml
-{
   masking: "*",
   plaintext_length: 5,
   plaintext_side: "left"
-}
 ```
 
 | Viewpoint                  | Visible data            |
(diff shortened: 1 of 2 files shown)


acra/security-controls/masking/_index.md
```diff
@@ -2,3 +2,56 @@
 title: Masking
 bookCollapseSection: true
 ---
+
+# Masking
+
+Partial column encryption.
+Hides encrypted part from those who don't have right decryption keys.
+In case of decryption failure, encrypted data will be replaced with a placeholder.
```
ilammy, line 10: Well, you *did* say that it's short, but this is *way too* short of a description, in my opinion.

I believe you should go into more detail on what masking does and provide some use cases for which masking makes a good fit. Like, why would you even want to disclose anything to a user without the right decryption keys or authorization to access the data. (First of all, I don't think it's about whether *the user* has the keys or not, but rather if the AcraServer instance which the data is accessed through has the required keys.)

acra/security-controls/masking/_index.md
```diff
@@ -2,3 +2,56 @@
...
+* `reencrypting_to_acrablocks` — whether to automatically re-encrypt values stored
+  in [AcraStructs]({{< ref "acra/acra-in-depth/data-structures/_index.md#acrastruct" >}})
+  to [AcraBlocks]({{< ref "acra/acra-in-depth/data-structures/_index.md#acrablock" >}})
+<!-- TODO add link to page where colemn encryption settings are described in general, with client_id, zone_id etc -->
+
+<!-- Config validation func lives in masking/common/patterns.go -->
+In order to enable masking you will need:
+* non-empty value for `masking`
+* `plaintext_length` ⩾ 0
+* valid value for `plaintext_side` (`left` or `right`)
```
ilammy, line 36: I find the “commented example configuration” to be one of the best ways to document configuration files.

For example:

```
{
    // (required, must be non-empty)
    // Replacement string to use for masking out encrypted data.
    // If decryption fails, this string will be show instead of the masked portion of the data.
    "masking": "XXXXXXXXXXXX",

    // (required, must be non-negative)
    // How many bytes of plaintext to leave unencrypted.
    "plaintext_length": 4,

    // (required, allowed values: "left", "right")
    // Which side of the plaintext to leave unencrypted.
    // "plaintext_length" bytes will be retained from left or right side of the data cell.
    "plaintext_side": "right",

    // (optional, default: "acrastruct", allowed values: "acrastruct", "acrablock")
    // Which cryptographic container to use for data encryption.
    "crypto_envelope": "acrastruct",

    // (optional, default: false)
    // If true, data stored in AcraStructs will be transparently reencrypted into AcraBlocks.
    "reencrypting_to_acrablocks": false,
}
```

This also shows one of the reasons why JSON is not particularly suited for configuration files, since the spec does not include comments. Sure, *your* parser might support and ignore them but others might not. This creates unnecessary friction, for example, when you want to add syntax highlighting to your documentation but that parser does not want to recognize comments.

acra/security-controls/masking/_index.md
```diff
@@ -2,3 +2,56 @@
...
+
+## Examples
+
+Here are few examples of configuration and results it would give
+
+`{ masking: "*", plaintext_length: 5, plaintext_side: "left" }`
+* data, successful decryption: `Alan Turing`
+* encrypted: `Alan <encrypted>`
+* unsuccessful decryption: `Alan *`
```
ilammy, line 45: Examples can be given names so that it's easier to understand the intended effect. Tables can be used to make it easier to see what happens to data. Also, I'd actually use something that looks like encryption instead of `<encrypted>` to avoid making an impression that encrypted data is replaced with literally `<encrypted>` as a placeholder.

```markdown
    Retaining only the first five characters:

    ```json
    {
      "masking": "*",
      "plaintext_length": 5,
      "plaintext_side": "left"
    }
    ```

    | Viewpoint | Visible data |
    | --------- |:------------ |
    | database, encrypted | `Alan PGVuY3J5cHRlZD4=` |
    | authorized user, decrypted | `Alan Turing` |
    | unauthorized used, masked | `Alan *` |
```

Rendered:

> Retaining only the first five characters:
>
> ```json
> {
>   "masking": "*",
>   "plaintext_length": 5,
>   "plaintext_side": "left"
> }
> ```
> 
> | Viewpoint | Visible data |
> | --------- |:------------ |
> | database, encrypted | `Alan PGVuY3J5cHRlZD4=` |
> | authorized user, decrypted | `Alan Turing` |
> | unauthorized used, masked | `Alan *` |

See how table alignment makes it more clear that the first five characters – `Alan ` – are retained in plain.

acra/security-controls/masking/_index.md
storojs72: Agree. The introduction should be expanded a bit. Following use-cases can be mentioned:
- credit-card number masking (when part of number is shown in banking);
- email/phone number masking (in password recovery).

acra/security-controls/masking/_index.md
iamnotacake: Took your variant. But had to split it a bit because of horizontal scroll we don't want. And made it YAML because we actually use this format for configuration.

> the spec does not include comments

That's why someone invented JSON5 :)

ilammy: > made it YAML because we actually use this format for configuration

I guess I got bamboozled by the provided examples:

```
{ masking: "*", plaintext_length: 5, plaintext_side: "left" }
```

which certainly looks JSON-like to me.

Meanwhile, I don't think that

```yaml
{
  masking: "???",
  plaintext_length: 0,
  plaintext_side: "left"
}
```

is valid YAML.

Shouldn't it be something like this instead?

```yaml
masking: "???"
plaintext_length: 0
plaintext_side: "left"
```

acra/security-controls/masking/_index.md
```diff
@@ -2,3 +2,122 @@
...
+  plaintext_length: 4,
+  plaintext_side: "right"
+}
+```
+
+| Viewpoint                  | Visible data            |
+| -------------------------- | :---------------------- |
+| database, encrypted        | `PGVuY3J5cHRlZD4=scal`  |
+| authorized user, decrypted | `Blaise Pascal`         |
+| unauthorized used, masked  | `XXXXscal`              |
```
ilammy, line 91: I believe just one example is enough. “Retain last four characters” looks like a perfect fit for masking CC digits, so I'd remove this example with names and leave only the following one.

acra/security-controls/masking/_index.md
iamnotacake: Every JSON is valid YAML AFAIK. This one seems valid too, checked on https://yamlvalidator.com/.
But your words make sense, guess it's better to remove `{` `}` OR make it look like a part of column encryption settings:
```yaml
# in config description

schemas:
  - table: foo
    encrypted:
      - column: first_name
        # description
        masking: "XXXX"
        # description
        plaintext_length: 1
        # description
        plaintext_side: "left"
        # same with crypto_envelope and reencrypting_to_acrablocks

# in examples

masking: "*"
plaintext_length: 1
plaintext_side: "left"
```

acra/security-controls/masking/_index.md
```diff
@@ -2,3 +2,122 @@
...
+    reencrypting_to_acrablocks: false,
+}
+```
+<!-- TODO add link to page where colemn encryption settings are described in general, with client_id, zone_id etc -->
+
+## Examples
+
+Here are few examples of configuration and results it would give.
+The actual encrypted part of data will look differently in database, and will take more bytes.
+Examples below only ilustrate which part of plaintext would be encrypted.
```
Lagovas, line 59:
```suggestion
Examples below only illustrate which part of plaintext would be encrypted.
```

acra/security-controls/masking/_index.md
ilammy: > Every JSON is valid YAML AFAIK.

TIL.

How can this world be *more* messed up? 🤔

Lagovas (approved):
lgtm

PR #71, opened 2021-09-01
Secure logging

Lagovas:
Here described our secure logging and acra-log-verifier.
Part of description about logging I took from our internal wiki written by @storojs72 and @ilammy.

Additionally migrated `contribution` page from [old site](https://docs.cossacklabs.com/pages/documentation-acra/#contributing-to-acra).

### DIFF
--- a/acra/configuring-maintaining/general-configuration/acra_log_verifier.md
+++ b/acra/configuring-maintaining/general-configuration/acra_log_verifier.md
@@ -1,0 +1,293 @@
+---
+title: AcraLogVerifier
+bookCollapseSection: true
+weight: 3
+---
+
+# AcraLogVerifier
+
+AcraLogVerifier is command-line utility that verifies secure logs dumped from AcraServer/AcraTranslator/AcraConnector started with `--audit_log_enable=true` flag.
+
+It expects symmetric key to decrypt keys from keystore from  `ACRA_MASTER_KEY` environment variable in pair `--keys_dir` flag or from HashiCorp Vault together with `--vault_*` flags. 
+
+## Command line flags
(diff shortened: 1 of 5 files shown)


acra/configuring-maintaining/general-configuration/acra_log_verifier.md
```diff
@@ -0,0 +1,293 @@
...
+  and will not interrupt validation with non-zero exit status.
+
+* `-d`
+
+  Debug mode, shows order of input files. Useful in pair with `--audit_log_file_list` flag. 
+  Verifier will print absolute file paths that in the order used during validation. 
+
+### Logging
+
+* `--logging_format={plaintext|json|CEF}`
```
ZhmakAS, line 48: Do we really need to duplicate such information, I remember adding some logging info here:
https://github.com/cossacklabs/product-docs/blob/master/acra/configuring-maintaining/logging/_index.md

Lagovas: We have already existing structure of CLI description for our tools with common sections, logging also too. And actually, I missed already existing section in `logging` ) but I think it's okay to duplicate. On separate page we can describe it more verbosely. On CLI pages it more useful when all arguments listed together with brief description. Plus `acra_log_verifier` may use these flags in another way. Especially here this flag specify type of log file that will be used, not logs that will be generated.

ZhmakAS: Agree, makes sense

ZhmakAS (approved):
lgtm

PR #78, opened 2021-09-16
acra-webconfig description

Lagovas:
Here described our acra-webconfig service. Plus added screenshots of current state of UI.

PR #79, opened 2021-09-16
Describe acra-authmanager

Lagovas:
There is CLI description for `acra-authmanager` with example of auth.file. Created empty file for `acra-webconfig` to refer to this page at this PR and not break hugo rendering. Content of that page will be overwritten by #78.

### DIFF
--- a/acra/configuring-maintaining/general-configuration/acra-authmanager.md
+++ b/acra/configuring-maintaining/general-configuration/acra-authmanager.md
@@ -1,0 +1,131 @@
+---
+title: acra-authmanager
+bookCollapseSection: true
+---
+
+# acra-authmanager
+
+`acra-authmanager` is CLI utility for [acra-webconfig](/acra/configuring-maintaining/general-configuration/acra-webconfig.md) user management.
+Using this utility you can add/update/remove users that should have access to web UI of `acra-webconfig`. It changes 
+encrypted auth file that stores all users and hashed passwords for basic authentication.
+
+## Command line flags
+
+`acra-authamanager` should be called with same `ACRA_MASTER_KEY` that used for `acra-server`. Only these two binaries can
+read encrypted file with authentication data. 
+
+### Configuration files
+
+* `--file=<filepath>`
+  
+    Path to encrypted file where stored authentication data
+    Default is `configs/auth.keys`
+
+* `--user=<username>`
+
+    Name of user that will be added/removed/updated. This flag is required.
+  
+* `--password=<password>`
+
+    User's password that will be used for basic authentication. Cannot be empty for `--set` operation and unused for 
+    `--remove` operation.
+  
+* `--remove`
+  
+    Flag to remove user specified by `--user=<user>` flag. 
+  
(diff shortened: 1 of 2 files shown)


acra/configuring-maintaining/general-configuration/acra-authmanager.md
```diff
@@ -0,0 +1,131 @@
+---
+title: acra-authmanager
```
iamnotacake, line 2: In some places (inside  acra/configuring-maintaining/general-configuration/*.md) we use `AcraFoo`, in others — `acra-foo`.
Just a note that we'll have to end up with a single variant.

acra/configuring-maintaining/general-configuration/acra-authmanager.md
```diff
@@ -0,0 +1,131 @@
...
+bookCollapseSection: true
+---
+
+# acra-authmanager
+
+`acra-authmanager` is CLI utility for [acra-webconfig](/acra/configuring-maintaining/general-configuration/acra-webconfig.md) user management.
```
iamnotacake, line 8: What about using another form? If it makes actual difference of course.
```suggestion
`acra-authmanager` is CLI utility for [acra-webconfig]({{< ref "/acra/configuring-maintaining/general-configuration/acra-webconfig.md" >}}) user management.
```

acra/configuring-maintaining/general-configuration/acra-authmanager.md
```diff
@@ -0,0 +1,131 @@
...
+`acra-authmanager` updates encrypted file where stored rows in format: 
+```
+<user>:<salt>:<hash_function_parameters>:<hash(salt, password, parameters)>
+<user>:<salt>:<hash_function_parameters>:<hash(salt, password, parameters)>
+<user>:<salt>:<hash_function_parameters>:<hash(salt, password, parameters)>
+``` 
+Each row is separate entry related to distinct user. 
+
+Every password hashed using [Argon2 hash function](https://en.wikipedia.org/wiki/Argon2). You can find current parameter
+values used for hash function in [Acra source code](https://github.com/cossacklabs/acra/blob/master/cmd/constants.go#L34).
```
iamnotacake, line 122: This link may become a bit outdated at some point.
Another possible solution would be using search link like [this one](https://github.com/cossacklabs/acra/search?q=DefaultWebConfigAuthArgon2Length), although it also points to other usages of the constant.

iamnotacake (approved):
Looks good, no obvious problems

acra/configuring-maintaining/general-configuration/acra-authmanager.md
Lagovas: changed to link with specified release number that will not be changed

acra/configuring-maintaining/general-configuration/acra-authmanager.md
Lagovas: thanks, I just missed. copied from IDE and forgot to wrap with tag

PR #81, opened 2021-09-16
Added acra_backup docs

ZhmakAS:
Added `acra_backup` cli tools docs

### DIFF
--- a/acra/configuring-maintaining/general-configuration/acra-backup.md
+++ b/acra/configuring-maintaining/general-configuration/acra-backup.md
@@ -1,0 +1,132 @@
+---
+title: acra-backup
+bookCollapseSection: true
+weight: 10
+---
+
+# acra-backup
+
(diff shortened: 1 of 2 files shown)


acra/configuring-maintaining/general-configuration/acra-backup.md
```diff
@@ -0,0 +1,132 @@
+---
+title: acra-backup
+bookCollapseSection: true
+weight: 10
+---
+
+# acra-backup
+
+`acra-backup` is a command-line utility used for storing and managing the keystore [backups]({{< ref "/acra/keys/operations/backup" >}}).
```
Lagovas, line 9: would be nice to mention that with acra-backup users can migrate keys from one keystore to another, or to another storage type. for example from filesystem storage to redis and vice versa. and would be great to add examples with such manipulations. for example export keys from folder with keys and import to redis storage.

ZhmakAS: Agree, nice to have

Lagovas (approved):
lgtm

PR #83, opened 2021-09-18
What is Acra, v1

9gunpi:
This is rough outline of Getting started/What is Acra section. It will be expanded, I'm just trying things out and sketching the plan, but feel free to comment.


vixentael (approved):
thank you for your hard work, @9gunpi! todos received `o7`

PR #85, opened 2021-09-21
Cryptography in acra

Lagovas:
Migrated this page - https://docs.cossacklabs.com/pages/documentation-acra/#cryptography-in-acra
A bit updated with cryptography usage in acra and updated mentioned services that uses cryptography, because was mentioned only AcraServer and AcraConnector)

And I'm not sure do we need mention Acra Enterprise with support of certified crypto. But mentioned links with prices works now - https://www.cossacklabs.com/acra/#pricing . May be @vixentael or @9gunpi can clarify.


ZhmakAS (approved):
Lgtm

PR #89, opened 2021-09-22
Add acra-keys docs

ZhmakAS:
`acra-keys` docs cli description with all sub-commands


acra/configuring-maintaining/general-configuration/acra-keys/export.md
```diff
@@ -0,0 +1,120 @@
...
+
+**`export`** is `acra-keys` subcommand used for exporting keys of the keystore version `v2`.
+
+## Command line flags
+
+### General flags
+
+* `--key_bundle_file=<path>` ❗
+
+  Path to input file for exported key bundle.
```
Lagovas, line 17: path to input or output file? Do we export into this file or load something from there? in first case, imho, we should use `path to output file`

acra/configuring-maintaining/general-configuration/acra-keys/extract-client-id.md
```diff
@@ -0,0 +1,49 @@
...
+
+  ❗ - flags required to be specified.
+
+
+## Usage example
+
+**`extract-client-id`** subcommand could be used to generate ClientID to reference some keys generated by [`acra-keymaker`][**`export`** subcommand]({{< ref "/acra/configuring-maintaining/general-configuration/acra-keymaker" >}}) via `tls_cert` flag and empty `client_id`.
+
+```
+$ acra-keys extract-client-id --tls_cert=/etc/acra-demo/configs/acra-client1.crt --print_json
```
Lagovas, line 39: would be great to add example of output without `--print_json` flag. it will simplify understanding what to expect and how it can be used in another case.

acra/configuring-maintaining/general-configuration/acra-keys/generate.md
```diff
@@ -0,0 +1,194 @@
...
+* `--acrawebconfig_symmetric_key`
+
+  Generate a new symmetric key for encrypting AcraWebconfig's basic authentication credentials.
+
+
+### Zone keys
+
+* `--zone`
+
+  Generate new Acra storage zone.
```
Lagovas, line 184: it is not obvious flag description that pushed me to read source code to understand what actually it doing)
`--zone` generates new Zone keypair (without symmetric key). and `--zone_storage_key` rotates existing by generating new and changing current to read-only state, for decryption purposes only.

acra/configuring-maintaining/general-configuration/acra-keys/export.md
ZhmakAS: Yes, you are right, my typo

Lagovas (approved):

PR #101, opened 2021-09-29
Add content for horizontal/vertical scaling

iamnotacake:
Add content for
* Configuring & maintaining > Optimizations > Horizontal scaling
* Configuring & maintaining > Optimizations > Vertical scaling

based on data from T1956, add few hints regarding AcraServer banchmarking


acra/configuring-maintaining/optimizations/vertical_scaling.md
```diff
@@ -5,4 +5,38 @@ weight: 1
...
+|       |          |        |
+| 8     | 1        | 238M   |
+| 8     | 8        | 567M   |
+| 8     | 20       | 746M   |
+
+\* — number of connections handled by AcraServer in parallel,
+each producing hundreds or thousands of request one after another
+
+It is recommended to give the machine at least twice the amount you see in the table,
+900M for single core machine, 1200M for dual core one and so on.
```
shadinua, line 42: I managed to roughly sketch out the formula for calculating the amount of the minimum recommended RAM, now it looks like this:
```
RAM = 263.83 + connections × 6.8 + cores × 43.57
```
I've created a [table with calculated data and quickly built a chart](https://docs.google.com/spreadsheets/d/1J-l2A4n7YR3hd7XXTLDoW4QYyQ0stBvJQ4CeG5fInQ4/edit?usp=sharing), which I'd suggest to add here instead of text.

So this chapter could be turn into something like:
```suggestion
Here are some rough recommendations for minimum memory amount based on our tests:

<PUT_IMAGE_HERE>

where:
  "connections" — number of connections handled by AcraServer in parallel
  "cores" — number of CPU cores available to AcraServer
  
This recommendations might be used only for an approximate calculation. The actual use of RAM can vary depending on many factors, for example, the nature of the load, the number and type of client keys, the type of database, etc. 
```

shadinua (approved):
Great job, thanks!

PR #96, opened 2021-09-23
Security controls > Transport security

iamnotacake:
Fill pages that describe AcraConnector and TLS as transport encryption.
Mention AcraWriter that can encrypt data on client side. Not sure about its name (called _separate encryption_) though.

Fix few things noticed during docs exploration.

PR #102, opened 2021-09-29
ToCs for many docs, placeholders, plans, etc.

9gunpi:
This features many missing docs' table of contents and suggestions. Please troubleshoot them for conflicts, but do not try to resolve links RN.

### DIFF
--- a/acra/acra-in-depth/architecture/_index.md
+++ b/acra/acra-in-depth/architecture/_index.md
@@ -1,4 +1,37 @@
 ---
 title: Architecture
 bookCollapseSection: true
+weight: 1
 ---
+
+# Architecture
+
+## Key components
+
+Security control services: Proxy and API 
+Key storage
+Configuration / policy
+KMS
+
+## Supporting modules
+
+AcraConnector
+WebConfig
+
+
+# Data flow & component layout
+
+
+Acra requires 3 components to talk together during startup: 
+
+* KMS to fetch keys 
+* Acraserver/AcraTranslator
+
+Acra requires 3 components to talk together during encryption/decryption: 
+
+* Key store
+* AcraServer/Translator
+* Application
+* Database (in AcraServer case)
+
+For examples, please consult (TODO link) appropriate section for data flow examples. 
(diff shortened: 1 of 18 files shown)


acra/acra-in-depth/architecture/_index.md
```diff
@@ -1,4 +1,37 @@
 ---
 title: Architecture
 bookCollapseSection: true
+weight: 1
 ---
+
```
Lagovas, line 6:
```suggestion
**TODO: Expand**
```
Will be great to have some marker that we can `grep` and find where we should do something.

acra/security-controls/transport-security/_index.md
```diff
@@ -2,3 +2,6 @@
 title: Transport security
 bookCollapseSection: true
 ---
+
```
Lagovas, line 5:
```suggestion
**TODO: Expand**
```
Some marker to `grep` before publishing. But will be better to add comments in the appropriate PR #96 and @iamnotacake can apply before merging.

acra/acra-in-depth/architecture/_index.md
9gunpi: This is indeed intended marker for grep,- I did grep the whole documentation repo before starting to use it and I think TODO (ALLCAPS) is not used anywhere for reasons other than actual todo. 

Moreover, I believe, **TODO: Expand** is unique enough to grep for it.

acra/security-controls/transport-security/_index.md
9gunpi: Same here - this is indeed marker for grep. 

If you want it to be more explicit - suggest ;)

acra/acra-in-depth/architecture/_index.md
Lagovas: I suggested adding this line here because it was missed here and in the `transport-security/_index.md` files. and exists in all other docs that you updated. Yes, we used TODO as marker. And because this doc file is not finished, it will be good to add `TODO` line too

9gunpi:
Please review, I've accepted your suggestions

Lagovas (approved):
lgtm, but PR has merge conflicts. probably due to merged other PRs

PR #103, opened 2021-10-01
Key storing v1

Lagovas:
There is the first version of the content for KMS and key-storing pages.
I'm not sure how we want to describe our KMS support, probably somehow tell that we can extend KMS integrations after buying an enterprise license.
I didn't start to write content for `kv-stores.md` with description of supported storages because I'm not sure that I can finish. Plus we should decide, do we want to describe there about all storages that we support or only keys related. Because we don't support boltdb and in-memory (only in-memory cache) storages for keys, only for tokens.

### DIFF
--- a/acra/configuring-maintaining/key-storing/_index.md
+++ b/acra/configuring-maintaining/key-storing/_index.md
@@ -4,6 +4,43 @@ bookCollapseSection: true
 weight: 1
 ---
 
-**TODO: Expand**
-
 # Key management and key storage in general 
+
+Key management with key storage is core component of Acra responsible for security guarantees. The whole cryptography 
+used in Acra depends on key storage. Without great key management, even strong and correctly used cryptography can't 
+guarantee safety for data.
+
+Key management is a set of approaches and processes related to key storing, rotation, distribution, protection, 
+compartmentalization. Acra provides tools and pushes the processes in key management in the secure way to cover all 
+aspects of keys security and help to reach compliance in data security.
+
+## Keys protection
+
+Every private key that stored and used by Acra encrypted with [Secure Cell]({{< ref "/themis/crypto-theory/cryptosystems/secure-cell.md" >}})
+that designed and responsible for secure storage. To encrypt all private keys Acra needs master key for symmetric cryptography
+under the hood of Secure Cell. This master key may be passed to Acra in two ways:
+* environment variable
+* KMS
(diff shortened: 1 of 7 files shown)


vixentael:
> Plus we should decide, do we want to describe there about all storages that we support or only keys related

I think we should start from something. I suggest shortly describing kv-stores for keys and for tokens.

ZhmakAS (approved):

acra/configuring-maintaining/key-storing/_index.md
```diff
@@ -4,6 +4,43 @@ bookCollapseSection: true
...
+used in Acra depends on key storage. Without great key management, even strong and correctly used cryptography can't 
+guarantee safety for data.
+
+Key management is a set of approaches and processes related to key storing, rotation, distribution, protection, 
+compartmentalization. Acra provides tools and pushes the processes in key management in the secure way to cover all 
+aspects of keys security and help to reach compliance in data security.
+
+## Keys protection
+
+Every private key that stored and used by Acra encrypted with [Secure Cell]({{< ref "/themis/crypto-theory/cryptosystems/secure-cell.md" >}})
```
iamnotacake, line 19:
```suggestion
Every private key that is stored and used by Acra is encrypted with [Secure Cell]({{< ref "/themis/crypto-theory/cryptosystems/secure-cell.md" >}})
```

acra/configuring-maintaining/key-storing/_index.md
```diff
@@ -4,6 +4,43 @@ bookCollapseSection: true
...
+that designed and responsible for secure storage. To encrypt all private keys Acra needs master key for symmetric cryptography
+under the hood of Secure Cell. This master key may be passed to Acra in two ways:
+* environment variable
+* KMS
+  Passing master key via environment variable as simple as possible. You can find example on our page
+  [Necessary preparations]({{< ref "/acra/configuring-maintaining/general-configuration/necessary_prep.md" >}}).
```
iamnotacake, line 25: I guess we don't need that indentation, otherwise this master key description may be rendered right next to "KMS", not in the next line as you probably expected

```suggestion

Passing master key via environment variable as simple as possible. You can find example on our page
[Necessary preparations]({{< ref "/acra/configuring-maintaining/general-configuration/necessary_prep.md" >}}).
```

iamnotacake (approved):
Thanks. Approved

PR #124, opened 2021-10-18
Configuring & maintaining > Storage and data model implications

iamnotacake:
Define limitations introduced when Acra components are used for data protection

### DIFF
--- a/acra/configuring-maintaining/storage-and-data-model-implications/_index.md
+++ b/acra/configuring-maintaining/storage-and-data-model-implications/_index.md
@@ -8,7 +8,9 @@ bookCollapseSection: true
 Based on which Acra components you use, the requirements may differ.
 
 For example, you will most likely want to use AcraServer.
-In this case, you are limited by supported DBMS, which are MySQL, MariaDB, PostgreSQL.
+In this case, you are limited by DBMS that use [MySQL](https://dev.mysql.com/doc/internals/en/client-server-protocol.html)
+or [PostgreSQL](https://www.postgresql.org/docs/current/protocol.html) wire protocols.
+We regularly test MySQL, MariaDB, PostgreSQL to make sure AcraServer works properly with them.
 
 On the other hand, there is AcraTranslator and it does not depend on storage you use.
 It's just a side service that performs crypto operations and you are free to store encrypted data in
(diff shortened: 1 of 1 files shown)


acra/configuring-maintaining/storage-and-data-model-implications/_index.md
```diff
@@ -2,3 +2,79 @@
 title: Storage and data model implications
 bookCollapseSection: true
 ---
+
+# Data storage
+
+Based on which Acra components you use, the requirements may differ.
+
+For example, you will most likely want to use AcraServer.
+In this case, you are limited by supported DBMS, which are MySQL, MariaDB, PostgreSQL.
```
Lagovas, line 11: limited by DBMS that uses [MySQL](https://dev.mysql.com/doc/internals/en/client-server-protocol.html) and [PostgreSQL](https://www.postgresql.org/docs/current/protocol.html) wire protocols. But we should note that it regularly  tested only on postgresql/mysql/mariadb

acra/configuring-maintaining/storage-and-data-model-implications/_index.md
```diff
@@ -2,3 +2,79 @@
...
+It's just a side service that performs crypto operations and you are free to store encrypted data in
+any other database (including embedded ones), in any key/value storage and so on.
+The main requirement for a storage would be support of binary fields (value as array of bytes, blob, etc).
+But even in this case you can, for example, base64 encode encrypted data and store it as a string.
+
+# Type compatibility
+
+When encrypting, Acra takes a bytestring and returns a bytestring.
+So all the data you wanna protect should be transformed into a bytestring:
+1) text strings can be taken "as is"
```
Lagovas, line 23: actually, it is not true because DBMS that expects strings will validate encoding and will return errors to clients if strings in invalid encoding after encryption. It is supported only if database doesn't validate encoding (MySQL in some cases, but not Postgresql).
imho better to remove this line at all )

acra/configuring-maintaining/storage-and-data-model-implications/_index.md
```diff
@@ -2,3 +2,79 @@
...
+to protect from such attacks you can, for example, make all numbers the same length by prepending zero digits
+(like `%010d` for `printf`)
+
+The responsibility of performing these kind of conversions will be on the application.
+
+## One way to solve type problem
+
+Instead of making everything you wanna encrypt a byte string,
+you can use [tokenization]({{< ref "acra/security-controls/tokenization/_index.md" >}}).
+This will help keeping column types as before, but the values won't be stored as plaintext anymore.
```
Lagovas, line 39: actually, values will be stored as plaintext, but source values will be stored securely as ciphertext. Better to explain that source values will be stored securely and replaced for storage with new random values.

acra/configuring-maintaining/storage-and-data-model-implications/_index.md
```diff
@@ -2,3 +2,79 @@
...
+SELECT `number`, `string` FROM `records`
+WHERE `encrypted_string`="secret three";
+```
+
+While these won't:
+```sql
+-- For the database, `encrypted_string` is not "secret three",
+-- but some completely different array of bytes
+UPDATE `records` SET `string`="test"
+WHERE `encrypted_string`="secret three";
```
Lagovas, line 70: actually with searchable encryption it should work

PR #133, opened 2021-10-20
Acra in depth > Security features

iamnotacake:
Q: Should we put all the security controls into one of Proactive/Reactive/Detective groups?
Q: How much description should be there? As far as I remember, there already are some pages that list and briefly describe security controls.


vixentael (approved):
I'll fix wording a bit during my review

PR #190, opened 2021-11-04
Minor updates

iamnotacake:


vixentael (approved):
What do you think about removing the whole "necessary preparations" page?

I think "configuring & maintaining" is a wrong place for this info. We already have described how to setup AS/AT in the guides "integrating AS / AT".

Let's move information about ports into AS and AT guides, and remove the whole neccessary-preps page?

vixentael (approved):

PR #208, opened 2021-11-30
add "troubleshooting installation" page for themis

vixentael:
Based on the numerous ~weird questions~ inquiries from the community, we've created "troubleshooting installation" page for Themis.

Important to remember: all `Hx` headers are shown in the search, so it makes sense for the Troubleshooting page to contain headers that start with "Troubleshooting" to easily distinguish them in searchbox.

Screenshot 2021-11-30 at 16 31 29

Screenshot 2021-11-30 at 16 31 33

### DIFF
--- a/themis/debugging/cli-utilities.md
+++ b/themis/debugging/cli-utilities.md
@@ -1,6 +1,6 @@
 ---
-weight: 2
-title:  CLI utilities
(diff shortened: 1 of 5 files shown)


vixentael:
🥺👉👈 @shadinua

PR #221, opened 2021-12-20
sunsetting Themis Server and removing its mentions

vixentael:
as we sunsetting Themis Server (aka Themis Interactive Simulator), I removed its references from docs.


Lagovas (approved):
holy shit, how many mentions in docs

vixentael:
yes, Themis Server is a cool thing, that i pity that it takes so much efforts to maintain, but is not being used by others...

I still like the idea though, maybe later we will reincarnate it in a better shape.

PR #233, opened 2022-02-16
Added nonce keyword

ColliderBang:


themes/book/layouts/partials/docs/html-head.html
```diff
@@ -8,18 +8,18 @@
 <title>{{ partial "docs/title" . }}{{if eq (ne .Site.Title "") (not .IsHome)}} | {{ .Site.Title -}}{{end}}</title>
 
 {{- $manifest := resources.Get "manifest.json" | resources.ExecuteAsTemplate "manifest.json" . }}
-<link rel="manifest"rel="manifest" crossorigin="use-credentials" href="{{ $manifest.RelPermalink }}">
-<link rel="icon" href="{{ "favicon.png" | relURL }}" type="image/x-icon">
+<link nonce="d8cc69c4355f969e" rel="manifest"rel="manifest" crossorigin="use-credentials" href="{{ $manifest.RelPermalink }}">
```
Lagovas, line 11: so how people should calculate this nonces if they want to update something? can we use some hugo tags for that? may be some hash function and pass link as payload. then it will be automated. plus I see below some `integrity` attributes. can we use it as nonce? how they calculated and what format do they have?

vixentael: the nonce are calculated as sha256 (see T2468). the nonce should be unique for all scripts on the site. if the script was changed, no need to change the nonce. however, if someone suddenly changes the nonce —  CSP policy should be updated.

see details in T2468

vixentael (approved):

PR #239, opened 2022-03-23
Remove the mention of poison key auto-generation

G1gg1L3s:
https://github.com/cossacklabs/acra/pull/516 removes poison key generation during a query, so  this PR updates docs about that.


Lagovas (approved):
cool catch

PR #245, opened 2022-04-18
Replace v2 keystore with v1 in examples

G1gg1L3s:
This PR updates docs and changes suggestions and examples to use keystore v1 due to better performance.

### DIFF
--- a/content/acra/configuring-maintaining/general-configuration/acra-keys/list.md
+++ b/content/acra/configuring-maintaining/general-configuration/acra-keys/list.md
@@ -5,7 +5,7 @@ weight: 1
 
 # list
 
-**`list`** is `acra-keys` subcommand used for describing keys of the keystore version `v2`.
+**`list`** is `acra-keys` subcommand used for describing keys of the keystore versions `v1|v2`.
 
 ## Command line flags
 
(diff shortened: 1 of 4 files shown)


content/acra/configuring-maintaining/general-configuration/acra-keys/list.md
```diff
@@ -107,9 +107,10 @@ $ acra-keys list
 
 INFO[0000] Initializing ACRA_MASTER_KEY loader...       
 INFO[0000] Initialized default env ACRA_MASTER_KEY loader 
-Key purpose                  | Client/Zone ID | Key ID
------------------------------+----------------+-------------------------
-audit log signature key      |                | audit-log
-client storage symmetric key | user1          | client/user1/storage-sym
+Key purpose     | Client/Zone ID | Key ID
+----------------+----------------+------------------
+audit_log       | secure         | secure_log_key
```
Lagovas, line 112: what means `secure` and where user should get it?

content/acra/configuring-maintaining/general-configuration/acra-keys/list.md
```diff
@@ -107,9 +107,10 @@ $ acra-keys list
...
+storage_sym_key | user1          | user1_storage_sym
+audit_log       | secure         | secure_log_key
+storage_sym_key | user1          | user1_storage_sym
```
vixentael, line 115: storage_sym_key twice?

vixentael (commented):
Do we have a description somewhere of the difference between v1 and v2?

Lagovas:
> Do we have a description somewhere of the difference between v1 and v2?

https://docs.cossacklabs.com/acra/security-controls/key-management/versions/
@vixentael

content/acra/configuring-maintaining/general-configuration/acra-keys/list.md
G1gg1L3s: I have no clue why there is a `secure` and why `sym_key` is printed twice :) I just run the examples to be sure they work and copy the output

Lagovas: hm... it is first part of filename: `secure_log_key`... - https://github.com/cossacklabs/acra/blob/master/keystore/filesystem/server_keystore.go#L927

but I don't know why we print same type of keys twice...
Anyway it is wrong output. I suggest to write in docs correct and expected output and create new task for futher investigation why it works so and fix. What do you think?

G1gg1L3s: Ok. But since I don't know how real records should look like, I would keep everything as it was

Lagovas (approved):

vixentael (approved):

PR #253, opened 2022-07-04
Update info about HTTP API

G1gg1L3s:
Because of https://github.com/cossacklabs/acra/pull/550 we have a new cli flag and a couple of limitations less. This PR describes it.

I've checked, but it may not be the all places that should be updated. If you remember something else, please let me know.

Merge this PR after the https://github.com/cossacklabs/acra/pull/550.


content/acra/configuring-maintaining/general-configuration/acra-server/_index.md
```diff
@@ -249,6 +249,17 @@ weight: 3
...
 
+* `--http_api_use_tls={true|false}`
+
+  Enables the TLS for the HTTP API. Use together with the `--http_api_enable=true`.  Uses the same key and certificates as the Acra proxy.
+
+  Default is `false`.
+
+  {{< hint warning >}}
+  **Note:**
+  The default configuration with `--http_api_use_tls=false` is **insecure**. Enable TLS whenever possible.
```
vixentael, line 260: then why default is false? :)

G1gg1L3s: Backward compatibility ☹️
I guess, in the future we will make it `true` by default. However it's not that straightforward, because the clients should specify TLS keys and certificates, which they don't always do (like in situations when they play with the database and acra locally, so there is no need of specifying the TLS params).

vixentael: okay.jpg

vixentael (approved):

vixentael:
@Lagovas will you take a look, or should we merge?

content/acra/configuring-maintaining/general-configuration/acra-server/_index.md
Lagovas: hm. looked one more time and see http_api_use_tls and http_api_enable... and I think that we should rename it to something like http_api_tls_transport_enable... I suggested it @G1gg1L3s in Slack and will repeat it in #550. So lets merge this PR after merge into acra repo.

vixentael (approved):

Lagovas (approved):

PR #267, opened 2022-09-13
Deprecate zones

Lagovas:
Added deprecation notes to all places where zones are mentioned. Additionally mentioned for AT that zoneID args will be removed and ignored.
Also deprecated acra-addzone.


content/acra/configuring-maintaining/general-configuration/acra-addzone.md
```diff
@@ -224,6 +224,25 @@ Should be provided only with `--keystore_encryption_type=<kms_encrypted_master_k
 {{< /hint >}}
 
 
+### KMS
```
Zhaars, line 227: Hm, but there is already the KMS block above. Why do we need another one?

content/acra/configuring-maintaining/general-configuration/acra-backup.md
```diff
@@ -108,6 +108,25 @@ Should be provided only with `--keystore_encryption_type=<kms_encrypted_master_k
 {{< /hint >}}
 
 
+### KMS
```
Zhaars, line 111: The same here. Maybe it's some git collision.

vixentael:
why this PR has some many commits? 🤔

content/acra/configuring-maintaining/general-configuration/acra-backup.md
Lagovas: i don't how and why I copied it... but i removed

Lagovas:
> why this PR has some many commits? thinking

i don't know. locally, I forked new branch from staging (didn't notice that current branch staging instead of master) I committed changes and after that merged origin/master... if it is important or simplify reviewing, I can force re-push.

Zhaars:
> > why this PR has some many commits? thinking
> 
> i don't know. locally, I forked new branch from staging (didn't notice that current branch staging instead of master) I committed changes and after that merged origin/master... if it is important or simplify reviewing, I can force re-push.

I think its ok, just need to squash before merging.

content/acra/acra-in-depth/security-design/threat-models-and-guarantees.md
```diff
@@ -80,6 +84,11 @@ By differentiating the sensitive data via [Zones](/acra/security-controls/zones/
 
 - Zone IDs are used as context when encrypting data using Themis Secure Cell, providing additional protection.
 
+{{< hint warning >}}
+Zones are deprecated since 0.94.0, will be removed in 0.95.0. 
+AcraEnterprise provides approach of differentiating ClientID dynamically via SQL.
```
vixentael, line 89:
```suggestion
[Acra Enterprise Edition](https://docs.cossacklabs.com/acra/enterprise-edition/) provides a better approach of differentiating ClientID dynamically via SQL.
```

vixentael: can we write couple of words about "differentiating ClientID dynamically" somewhere in the docs? Describe how it works, what it requires, and point to the [Acra Enterprise Edition](https://docs.cossacklabs.com/acra/enterprise-edition/).

content/acra/configuring-maintaining/general-configuration/acra-addzone.md
```diff
@@ -3,7 +3,7 @@ title: acra-addzone
 weight: 10
 ---
 
-# acra-addzone
+# acra-addzone (deprecated since 0.94.0)
 
 `acra-addzone` is a command-line utility that generates new [Zone keys](/acra/security-controls/zones/) for AcraBlocks/AcraStructs.
```
vixentael, line 8: add couple of words about differentiating ClientID dynamically via SQL?

vixentael (changes requested):
I suggest we add somewhere about differentiating ClientID dynamically via SQL. Which page would be the best?

Then, we should mention it as a feature on [Acra Enterprise Edition](https://docs.cossacklabs.com/acra/enterprise-edition/) page.

vixentael:
> I committed changes and after that merged origin/master... if it is important or simplify reviewing, I can force re-push.

Technically speaking, we shouldn't have huge differences between master and staging. But to avoid regrets, I suggest when merging you ensure that the changes are only yours (no sudden pieces as result of merging).

content/acra/acra-in-depth/security-design/threat-models-and-guarantees.md
Zhaars: As for me, it makes sense to create a separate PR that contains only info about dynamic ClientID extracting and not mix in one PR.

vixentael: well, agreed, let's put it to a different PR. It's just that we have mentioned dynamic ClientID extracting here, so I suggested to add more details. I'm totally OK with doing it in a separate PR

content/acra/configuring-maintaining/general-configuration/acra-addzone.md
Lagovas: not sure that it is good place on acra-addzone page. Because it is small tool for zones generation. If we want to mention this feature, much better to do it on page about zones in general.

content/acra/acra-in-depth/security-design/threat-models-and-guarantees.md
Lagovas: Okay, lets add it in separate PR. I mentioned it here because there are several ways of deprecating zones. Remove this paragraph, mark as deprecated and mark + suggest a replacement. Because it looks like we describe that it is cool way to increase security with zones and then we deprecate it )) 

But we can just remove mentioning zones here.

vixentael: Well, I don't think that we should remove mentioning zones completely, at least until we don't have separate Acra docs for each version. Image that some of our customers already use Zones and old Acra. They might need info about zones as they use it now.

content/acra/configuring-maintaining/general-configuration/acra-addzone.md
vixentael: I agree, we should describe differentiating ClientID somewhere (a page of your choice), and then _mention_ it on pages relevant to Zones. 

Something like "Zones are deprecated since 0.94.0, but if you are looking for a flexible approach of differentiating clients, check the `differentiating ClientID dynamically via SQL` page".

mark + suggest a replacement approach you've described above

vixentael (approved):

vixentael:
we merge? we wait?

PR #296, opened 2023-01-27
Actualize stable with docs update

Zhaars:
- Actualize info about acra-keys destroy subcommand
- Added info about recently added TLS flags in acra-rotate/acra-rollback tools
- Added info about http_api_tls_transport_enable flag default value


vixentael (approved):

vixentael:
but please after merging take a look that the server actually has re-built the docs and we see new content

Lagovas (approved):

PR #297, opened 2023-02-03
Update supported golang versions

Lagovas:
Added note that we support only latest version of the golang compiler according to last decisions and changes


vixentael (approved):

vixentael:
Please merge to staging && prod!

PR #301, opened 2023-02-15
Update docs with acra keys list with rotated keys

Zhaars:
Added info about `acra-keys` `list` with rotated keys. RE https://github.com/cossacklabs/acra/pull/636


vixentael (approved):
@Lagovas take a look?

Lagovas (changes requested):
we discussed changes in the acra-list after reviewing public API in docs) and probably will change command a bit and then update doc one more time

plus discussed, that example with rotated keys should include several rotated keys and current encryption keys

Lagovas (approved):

PR #302, opened 2023-02-23
Add info about acra-keys destroy

Zhaars:
Added info regarding `acra-keys` `destroy` extension from [PR](https://github.com/cossacklabs/acra/pull/641)


vixentael (approved):
i don't have any comments about style. i'm handling this to @Lagovas for comments on content

PR #306, opened 2023-06-05
jsthemis: Describe issues with Nodejs and Openssl

G1gg1L3s:
Since we are going to start to support Openssl 3, we have to provide instructions on how to deal with Nodejs that has Openssl statically compiled.

I'm not sure whether we should add "Matching OpenSSL versions" section at the top or bottom. This is a big issue, so maybe users should see it right away when they load the page.

Also, I provided a couple of ways for fixing the issue. However, not all of them are tested in our CI, they were tested only manually a couple of times. So the question arises - should we provide them at all?

### DIFF
--- a/content/themis/languages/nodejs/installation.md
+++ b/content/themis/languages/nodejs/installation.md
@@ -117,7 +117,7 @@ $ node -e "console.log(process.versions['openssl'])"
 3.0.2+quic
 ```
 
-If the semantic versions are the same, you are good to go! If they differ, however, there are a few ways to resolve the situation, but there is no easy answer:
+If the semantic versions are the same, you are good to go! If they differ, however, there are a few ways to resolve the situation, but there is no easy answer. For instance, here is what we use while testing Themis in CI:
 
 1. Choose the Node.js version with the OpenSSL that matches with your systems' one or that has no OpenSSL built in (there may be versions of Node.js that use shared OpenSSL).
 
@@ -125,8 +125,10 @@ If the semantic versions are the same, you are good to go! If they differ, howev
 
 2. Try another distribution with OpenSSL that matches Node.js.
 
-3. Try to install OpenSSL version that matches your Node.js. Probably, it will require building OpenSSL from the sources. Then, you will have to rebuild [Themis and JsThemis from sources](#building-latest-version-from-source). Don't forget to [specify the path](../../../installation/installation-from-sources/#cryptographic-backends) to the new OpenSSL.
+Furthermore, here is how community solves this issue:
 
-4. You can try to build Themis and JsThemis from the sources with [Boringssl engine](../../../installation/installation-from-sources/#boringssl).
+1. Try to install OpenSSL version that matches your Node.js. Probably, it will require building OpenSSL from the sources. Then, you will have to rebuild [Themis and JsThemis from sources](#building-latest-version-from-source). Don't forget to [specify the path](../../../installation/installation-from-sources/#cryptographic-backends) to the new OpenSSL.
+
+2. You can try to build Themis and JsThemis from the sources with [Boringssl engine](../../../installation/installation-from-sources/#boringssl).
 
 However, none of these options are ideal because they can lock you to specific versions of software and disable the ability to update components and dependencies.
--- a/content/themis/languages/nodejs/installation.md
+++ b/content/themis/languages/nodejs/installation.md
@@ -13,6 +13,10 @@ Themis Core is available as a system package for Linux and macOS.
 Usually you want to install the stable package to benefit from automatic dependency management and security updates.
 However, you can also build and install the latest JsThemis from the source code.
 
+  {{< hint warning >}}
+  **Note**: Due to some issues with Openssl statically linked into Nodejs, for correct JsThemis operation, you need to ensure that the Nodejs and system Openssl versions match. Otherwise, you may encounter failures or crashes at runtime. Refer to [Matching Openssl versions](#matching-openssl-versions) for more details.
+  {{< /hint >}}
+
 ## Installing stable version on Linux
 
 The easiest way to install Themis on Linux is to use package repositories for your system.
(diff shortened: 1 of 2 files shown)


content/themis/languages/nodejs/installation.md
```diff
@@ -13,6 +13,10 @@ Themis Core is available as a system package for Linux and macOS.
 Usually you want to install the stable package to benefit from automatic dependency management and security updates.
 However, you can also build and install the latest JsThemis from the source code.
 
+  {{< hint warning >}}
+  **Note**: Due to some issues with OpenSSL statically linked into Nodejs, for correct JsThemis operation, you need to ensure that the Nodejs and system OpenSSL versions match. Otherwise, you may encounter failures or crashes at runtime. Refer to [Matching OpenSSL versions](#matching-openssl-versions) for more details.
+  {{< /hint >}}
```
vixentael, line 18: why not

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
     ```bash
     npm install /path/to/jsthemis.tgz
     ```
+
+## Matching OpenSSL versions
+
+JsThemis is a simple addon to Nodejs that provides bindings to the Themis library, which is itself installed as a shared library in the system. In turn, Themis depends on OpenSSL installed in the system.
```
vixentael, line 99:
```suggestion
JsThemis is a simple library to Node.js that provides bindings to the Themis Core library, which is itself installed as a shared library in the system. In turn, Themis Core depends on OpenSSL installed in the system.
```

Nodejs -> Node.js please fix everywhere

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
...
+
+However, [Nodejs often comes with its own OpenSSL version](https://github.com/nodejs/TSC/blob/main/OpenSSL-Strategy.md) included in the binary. This is a problem for JsThemis, because due to how linkage works, some OpenSSL functions will be linked directly from Nodejs and the others will come from the system's OpenSSL. If these OpenSSL versions are not the same, JsThemis will not work correctly or crash.
```
vixentael, line 101:
```suggestion
However, [Nodejs often comes with its own OpenSSL version](https://github.com/nodejs/TSC/blob/main/OpenSSL-Strategy.md) included in the binary. This is a problem for JsThemis, because due to how linkage works, some OpenSSL functions will be linked directly from Nodejs and the others will come from the system's OpenSSL. If these OpenSSL versions are not the same, JsThemis could become confused and crash.
```

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
...
+
+```
+                          +-----> Node openssl
+                          |
+jsthemis ----> libthemis -+
+                          |
+                          +-----> System openssl
+```
+
+For these reasons, before installing JsThemis, it is critical to ensure that the Nodejs and system OpenSSL versions are the same. To do so, you can check the output of these commands:
```
vixentael, line 111:
```suggestion
For these reasons, before installing JsThemis, it is critical to ensure that the Node.js and system OpenSSL versions are the same. To do so, you can check the output of these commands:
```

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
...
+```bash
+$ openssl version
+OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
+$ node -e "console.log(process.versions['openssl'])"
+3.0.2+quic
+```
+
+If the semantic versions are the same, you are good to go! If they differ, however, there are a few ways to resolve the situation, but there is no easy answer:
+
+1. Choose the Nodejs version with the OpenSSL that matches with your systems' one or that has no OpenSSL built in (there may be versions of Nodejs that use shared OpenSSL).
```
vixentael, line 122:
```suggestion
1. Choose the Node.js version with the OpenSSL that matches with your systems' one or that has no OpenSSL built in (there may be versions of Node.js that use shared OpenSSL).
```

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
...
+
+   For example, Ubuntu Jammy Jellyfish (22.04) has OpenSSL 3.0.2, so the only Nodejs versions suitable for it are v18.0 and v18.1. On the other hand, Ubuntu Focal Fossa (20.04) has OpenSSL 1.1.1, which is compatible with Node LTS v14 and v12. However, be aware that OpenSSL 1.1.1 [will be unsupported after September 2023](https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/).
+
+2. Try another distribution with OpenSSL that matches Nodejs.
```
vixentael, line 126:
```suggestion
2. Try another distribution with OpenSSL that matches Node.js.
```

content/themis/languages/nodejs/installation.md
```diff
@@ -89,3 +93,40 @@ you can manually build and install the latest version of Themis from source code
...
+
+3. Try to install OpenSSL version that matches your Nodejs. Probably, it will require building OpenSSL from the sources. Then, you will have to rebuild [Themis and JsThemis from sources](#building-latest-version-from-source). Don't forget to [specify the path](../../../installation/installation-from-sources/#cryptographic-backends) to the new OpenSSL.
```
vixentael, line 128:
```suggestion
3. Try to install OpenSSL version that matches your Node.js. Probably, it will require building OpenSSL from the sources. Then, you will have to rebuild [Themis and JsThemis from sources](#building-latest-version-from-source). Don't forget to [specify the path](../../../installation/installation-from-sources/#cryptographic-backends) to the new OpenSSL.
```

vixentael (changes requested):

vixentael:
> Also, I provided a couple of ways for fixing the issue. However, not all of them are tested in our CI, they were tested only manually a couple of times. So the question arises - should we provide them at all?

Yes, but maybe we can add smth like "These method is tested on Themis CI" and don't add the same line for other methods? Add something "Here is how community solves this issue: ".

vixentael:
Please review my comments, implement, update if needed, merge and merge further into master -> preview. Check how it looks. Then preview -> staging -> stable.

vixentael (approved):

PR #308, opened 2023-06-09
merge to stable

vixentael:


vixentael:
@G1gg1L3s once this is merged, please check doc server in ~1h to see that your Node.js changes are public (just to double check)

PR #310, opened 2023-06-22
Add ubuntu jammy to supported versions

G1gg1L3s:
As I understand, we don't delete the old distributions from the supported list, even if we don't directly test the new themis on them, don't we?

### DIFF
--- a/content/themis/installation/installation-from-packages.md
+++ b/content/themis/installation/installation-from-packages.md
@@ -22,6 +22,7 @@ Supported systems:
   - Ubuntu 16.04 LTS “Xenial Xerus”
   - Ubuntu 18.04 LTS “Bionic Beaver”
   - Ubuntu 20.04 LTS “Focal Fossa”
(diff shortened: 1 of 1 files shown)


Lagovas (commented):
by default, we declare only distros that we test and sure that they works. Some of them we can test longer than they are supported by maintainers, when it takes not much time to maintain from our side. Sometimes, we can leave distros that we know and sure that they work.
so, let's add new supported OS and remove not supported Ubuntu Xenial and Debian stretch (which we removed due to it's deprecated by distro maintainers and it requires much time from our side to fix dependency issues)

G1gg1L3s:
Okay, will do!

vixentael:
> remove not supported Ubuntu Xenial and Debian stretch

let's write smth like "Ubuntu Xenial — supported until Themis 0.15"

G1gg1L3s:
Something like this, @vixentael?

image

vixentael:
yes, thank you

PR #317, opened 2023-09-25
Update PyThemis installation instructions

iamnotacake:
Mention that old `pythemis_install` is deprecated, don't remove that part yet.

Add new section for more modern installation way
* Describe how to build/install a `.whl` package (virtualenv)
* Describe how to build/install a `.deb` package (Debian-like distros)
* Describe how to build/install an `.rpm` package (CentOS, RHEL, probably other Fedora-like)

Wheel building checked on `debian:bookworm` and `registry.access.redhat.com/ubi8/ubi:8.1` (kinda like CentOS/RHEL 8).
`.deb` building checked on `debian:bookworm`. `.rpm` building checked on `centos:7`.

### DIFF
--- a/content/themis/languages/python/installation.md
+++ b/content/themis/languages/python/installation.md
@@ -78,6 +78,58 @@ you can manually build and install the latest version of Themis from source code
  1. [Build and install Themis Core library](/themis/installation/installation-from-sources/)
     into your system.
 
+ 2. Create a "wheel" package
+
+    <!-- FIXME: For some reason (missing install tools in venv?) the build fails when venv is activated,
+                resolve this or if that's not an issue, just remove this comment -->
+    Run this command outside of virtual environment:
+    ```bash
+    make pythemis_make_wheel
+    ```
+
+    Result package filename will be printed at the end, like this:
+    ```
+    Result: src/wrappers/themis/python/dist/pythemis-0.14.0-py2.py3-none-any.whl
+    ```
+
+ 3. Install PyThemis in your virtual environment
+
+    Activate your virtual environment and run
+    ```bash
+    pip install src/wrappers/themis/python/dist/pythemis-0.14.0-py2.py3-none-any.whl
+    ```
+
+If the virtual environment already contained PyThemis,
+add `--force-reinstall` option to `pip install` to rewrite the previous version
+
+<!-- DRAFT FOR OS PACKAGES FOR PROPER GLOBAL INSTALLATION
+To install PyThemis globally, you need to create a system package first:
+```bash
+make pythemis_make_os_pkg
+```
+
+The result will be located at `src/wrappers/themis/python`.
+
+For Debian/Ubuntu it's a `.deb` file, install with `sudo dpkg -i <filename>`.
+
+*TODO: Add more package types*
+
+-->
+
+## Building latest version from source (deprecated)
+
+{{< hint warning >}}
+[PEP 668](https://peps.python.org/pep-0668/) changed the way Python packages are managed.
+It is now recommended to use `pip install` to only install packages into isolated virtual environments
+while installing things system-wide should be done using OS package manager (like `apt` or `rpm`).
+This means running `sudo make pythemis_install` is no longer recommended as it copies files
+into system-managed Python directory that is no longer maintained by `pip`.
+Consider building a `.whl` and installing it into your venv instead.
+{{< /hint >}}
+
+ 1. [Build and install Themis Core library](/themis/installation/installation-from-sources/)
+    into your system.
+
  2. Install PyThemis package from the source code:
 
     ```bash
--- a/content/themis/languages/python/installation.md
+++ b/content/themis/languages/python/installation.md
@@ -139,7 +139,7 @@ Alternatively, you could install PyThemis system-wide.
 [PEP 668](https://peps.python.org/pep-0668/) changed the way Python packages are managed.
 It is now recommended to use `pip install` to only install packages into isolated virtual environments
 while installing things system-wide should be done using OS package manager (like `apt` or `rpm`).
-This means running `sudo make pythemis_old_install` is no longer recommended as it copies files
+This means running `sudo make pythemis_install` is no longer recommended as it copies files
 into system-managed Python directory (like `/usr/lib/python3.X/site-packages`) that is no longer maintained by `pip`.
 Consider building a `.whl` and installing it into your venv instead.
 {{< /hint >}}
@@ -150,7 +150,7 @@ Consider building a `.whl` and installing it into your venv instead.
  2. Install PyThemis package from the source code:
 
     ```bash
-    sudo make pythemis_old_install
+    sudo make pythemis_install
     ```
 
     The package will be installed globally into the system.
@@ -161,7 +161,7 @@ Consider building a `.whl` and installing it into your venv instead.
     make prepare_tests_all test_python
     ```
 
-To remove PyThemis after using `pythemis_old_install`, run
+To remove PyThemis after using `pythemis_install`, run
 ```bash
 sudo pip uninstall --break-system-packages pythemis
 ```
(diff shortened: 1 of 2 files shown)


content/themis/languages/python/installation.md
```diff
@@ -75,6 +75,58 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
 If the stable package version does not suit your needs,
 you can manually build and install the latest version of Themis from source code.
 
+ 1. [Build and install Themis Core library](/themis/installation/installation-from-sources/)
+    into your system.
+
+ 2. Create a "wheel" package
+
+    <!-- FIXME: For some reason (missing install tools in venv?) the build fails when venv is activated,
```
vixentael, line 83: 🤔

content/themis/languages/python/installation.md
```diff
@@ -75,6 +75,58 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
...
+*TODO: Add more package types*
+
+-->
+
+## Building latest version from source (deprecated)
+
+{{< hint warning >}}
+[PEP 668](https://peps.python.org/pep-0668/) changed the way Python packages are managed.
+It is now recommended to use `pip install` to only install packages into isolated virtual environments
+while installing things system-wide should be done using OS package manager (like `apt` or `rpm`).
```
vixentael, line 124: omg

content/themis/languages/python/installation.md
```diff
@@ -143,3 +141,8 @@ Consider building a `.whl` and installing it into your venv instead.
     ```bash
     make prepare_tests_all test_python
     ```
+
+To remove PyThemis after using `pythemis_old_install`, run
+```bash
+sudo pip uninstall --break-system-packages pythemis
```
vixentael, line 147: 🤔

content/themis/languages/python/installation.md
iamnotacake: Made improvement in Makefile that ignores venv if it was activated, so building the wheel works now in such scenario. As long as there is python3-setuptools installed in system. Alternatively we could remove that hack and just say in docs "hey, you need no venv when packing PyThemis"

content/themis/languages/python/installation.md
iamnotacake: That's what we gotta do on Debian Bookworm (Python 3.11.2, pip 23.0.1) if first installed PyThemis system-wide "the old way" and then trying to remove it. Otherwise, without the suspicious option, enjoy the error:
```
$ sudo pip uninstall pythemis
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.
```
Same error for non-`sudo` command, it just wants to be used in virtual environment. Or use the OS package (that's not done yet)

content/themis/languages/python/installation.md
vixentael: the easier it is for the user — the better.
@Lagovas wdyt?

vixentael:
@Lagovas could you please take a look?

content/themis/languages/python/installation.md
```diff
@@ -75,13 +75,151 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
 If the stable package version does not suit your needs,
 you can manually build and install the latest version of Themis from source code.
 
+ 1. [Build and install Themis Core library](/themis/installation/installation-from-sources/)
+    into your system.
+
+ 2. Install Python dependencies
+
+    Depending on how you want to install PyThemis, there are different requirements.
```
Lagovas, line 83: Let's list the available approaches. One into the virtualenv and one system-wide installed as system package. Because further we describe steps how install dependencies, but not obvious for what. And why the user needs .whl. Usually developers installs packages with commands like `pip install` or `python setup.py` and they don't care what the files are generated. They interested in the package installed into their local or system environment. So we should clarify that we support 2 approaches. Local installation into venv and you need next dependencies, or installation into the system using system package manager and you need next dependencies.

content/themis/languages/python/installation.md
```diff
@@ -75,13 +75,151 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
...
+
+<!--
+      For CentOS, RHEL:
+
+      ```bash
+      sudo yum install python3-setuptools python3-wheel
+      ```
+-->
+
+    * Or install into virtualenv
```
Lagovas, line 103: You can use markdown viewer/editor in the pycharm or idea, or you can open .md file in the github to look how it will be rendered. We need valid markdown due to after that it will be rendered with hugo server that has own markdown parser. 
image

content/themis/languages/python/installation.md
```diff
@@ -75,13 +75,151 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
...
+      sudo gem install fpm
+      ```
+
+    On RHEL you also need to have working `pip` command. If it's not, and you only got `pip3`,
+    create a symlink like this: `sudo ln -s $(which pip3) /usr/bin/pip`.
+-->
+
+ 3. Create a PyThemis package
+
+    * Create a "wheel" for virtualenv
```
Lagovas, line 137: I prefer to explicitly note for which way this case. For example `To install locally: Create a "wheel"...`. Or mentions at start that we will describe 2 ways, first for local installation and second one for system. And then mention it like **Way 1. ....**, **Way 2. ....**

Because when I read step 3, I didn't understand why we need .whl for system package below and went to the makefile in the themis...

content/themis/languages/python/installation.md
```diff
@@ -75,13 +75,151 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
...
+
+<!--
+      For CentOS, RHEL:
+
+      ```bash
+      make pythemis_install_rpm
+      ```
+-->
+
+## Building latest version from source (deprecated)
```
Lagovas, line 205: I don't like this title for this block. As I understand, we need it for users who want install old version of the themis from the sources. And when they will pull old version, they will not have new makefile that knows how to pack .whl/.deb/.rpm.

@vixentael , what do you think, do we need instructions how to install old versions themis from the sources? If yes, how you suggest this title? Below we will describe how install pythemis from the sources themis < 0.16

Lagovas:
> @Lagovas could you please take a look?

I planned to look on this after merging https://github.com/cossacklabs/themis/pull/1023 that is source of instructions for this docs

content/themis/languages/python/installation.md
```diff
@@ -75,13 +75,146 @@ Once PyThemis is installed, you can [try out code examples](../examples/).
 If the stable package version does not suit your needs,
 you can manually build and install the latest version of Themis from source code.
 
+### Python wheel
```
Lagovas, line 78: last comment. let's mention in the paragraph above that there are 2 ways how to install from the sources, building and installing python package from wheel file and building os package in installing system-wide. 

Or just rephrase to `... you can manually build python wheel or .deb/.rpm package and install the latest version of Themis from source code.`

content/themis/languages/python/installation.md
vixentael: I'd leave this instruction for older Themis, but with note that it's for old Themis.

```suggestion
## Building Themis < 0.15 version from sources (deprecated)

{{< hint warning >}}
The instructions below are relevant only for Themis older than 0.15. Most likely, you don't need to follow them, and should install the latest Themis instead.
{{< /hint >}}
```

vixentael (changes requested):
👀

content/themis/languages/python/installation.md
iamnotacake: It's GitHub that renders like this. When running Hugo locally, it renders properly in web browser

content/themis/languages/python/installation.md
iamnotacake: Done

Lagovas (approved):

vixentael (approved):

PR #319, opened 2023-11-15
Add CPPFLAGS, LDFLAGS to installation example for jsthemis on macOS

radetsky:
I did something simple. I took an empty project and tried to install and use jsthemis. I remembered what the problem was, reread the sources, went through the history of correspondence and here I have prepared a recipe.


radetsky:
I've decided to extend the case after improving the bindings.gyp for macOS case. 
Most installation processes will be successful.

vixentael (approved):

PR #322, opened 2023-11-20
verified example of Dockerfile to fix nodejs jsthemis installation

radetsky:
Many thanks to @G1gg1L3s .

I installed it 10 times with Docker and 10 times on a live system using a cloud provider host and it worked 10/10.

### DIFF
--- a/content/themis/languages/nodejs/installation.md
+++ b/content/themis/languages/nodejs/installation.md
@@ -226,3 +226,64 @@ Error: Key Pair generation failed: invalid parameter
   code: 12
 }
 ```
+
+### Verified example how to fix it.
+You need to install `OpenSSL` and `Themis` native library from the source code and install the `jsthemis` package.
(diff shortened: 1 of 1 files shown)


content/themis/languages/nodejs/installation.md
```diff
@@ -226,3 +226,64 @@ Error: Key Pair generation failed: invalid parameter
   code: 12
 }
 ```
+
+### Verified example how to fix it.
```
Lagovas, line 230: what we are suggesting here? We show example of dockerfile that works. But our main goal is to describe how to build themis from the sources.  Why we shod dockerfile instead of commands for ubuntu/debian how to do it? Or if we rely on compilation using docker containers then let's describe how it use then. How to copy from the docker container and place into the target system

vixentael (changes requested):

PR #325, opened 2024-02-26
remove passphrase support from token protect/context imprint modes

Lagovas:
Our [user found](https://github.com/cossacklabs/themis/issues/1044) that we don't have passphrase support for secure cell token protect/context imprint mode and he is right. Let's synchronize docs with current implementation


vixentael (approved):

vixentael:
please merge to master - staging - stable @Lagovas

